PR #3, opened 2015-05-06
Add PGRES_COMMAND_OK to the check_status function to allows the create table query

werner:


will:
Nice catch. Can you please make it first `create table if not exists test()`, then ensure a drop after the spec so it can clean up. Right now it fails if you don't already have the table test

will:
Great thanks!

PR #7, opened 2015-05-29
Default extra_float_digits to 3 to preserve IEEE754 float fidelity

uhoh-itsmaciek:
The `extra_float_digits` setting must be at least `2` when reading 64-bit floats (and `3` for 32-bit floats) from Postgres to ensure that fidelity is preserved for all values (i.e., client and server are talking about the exact same bits).

As the Postgres community likes to point out, this is only meaningful for when both client and server have IEEE754 floating point implementations, but come on...

This is not ready to merge because of a spec failure, but I'm not sure if this is due to some issue in my patch, crystal-pg, or in crystal itself. The [semantically-equivalent test passes in pq](https://github.com/lib/pq/blob/master/conn_test.go#L845-L862). Any ideas?

### DIFF
--- a/spec/pg/connection_spec.cr
+++ b/spec/pg/connection_spec.cr
@@ -10,51 +10,51 @@ end
 
 describe PG::Connection, "#exec untyped" do
   it "returns a Result" do
-    res = DB.exec("select 1")
+    res = $DB.exec("select 1")
     res.class.should eq(PG::Result(Array(PG::PGValue)))
   end
 
   it "raises on bad queries" do
-    expect_raises(PG::ResultError) { DB.exec("select nocolumn from notable") }
+    expect_raises(PG::ResultError) { $DB.exec("select nocolumn from notable") }
   end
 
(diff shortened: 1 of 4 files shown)


will:
http://crystal-lang.org/docs/syntax_and_semantics/literals/floats.html
>  If no suffix is present, the literal's type is Float64.

so  you want `res.rows.should eq([{0.10000122_f32}])`

uhoh-itsmaciek:
Ah, cool. I'll fix this and move the specs to the rest of the encoding / decoding tests.

will:
Also you probably saw it since I mentioned you, but for cross-referencing fun, I opened an issue about the constant/global thing https://github.com/manastech/crystal/issues/708

uhoh-itsmaciek:
Rebased and force-pushed, fixed and broken up into neater commits.

Note that this breaks 8.4 and older, since they won't accept an extra_float_digits higher than 2.

PR #15, opened 2015-08-25
v0.3.0

will:
This was on master for a few hours, but I pulled it after discovering that under --release mode it broke.

Opened an issue https://github.com/manastech/crystal/issues/1274 but it's also likely that it's my fault.


will:
fixed in 0.3.1

PR #16, opened 2015-08-26
exec_all method to allow multiple requests

solisoft:
This simple method allow to run multiple requests at once.

Useful when loading SQL files (migrations)


will:
At first I thought that the first `#exec` method with no params or types would be good to also support multiple statements, but then it seems in libpq if you want the binary result format you have to use `PQexecParams` which doesn't allow multiple statements. There might be some way I'm not seeing though.

This is good for now thanks!

PR #18, opened 2015-09-05
Allows typed querying when to_hash method is used in query results

werner:
Hi, I tried to use:
```crystal
    DB.exec({String, String, Bool, Int32},
                  "select 'a' as foo, 'b' as bar, true as baz, 10 as uhh
                   union all
                   select '', 'c', false, 20").to_hash
```

Gives the next error:
```
instantiating 'PG::Result({Int32:Class, String:Class, String:Class, String:Class})#to_hash()'

    connect {|db| user = db.exec({Int32, String, String, String}, 
                                          ^~~~~~~

in ./libs/pg/pg/result.cr:37: no overload matches 'Hash(K, V)::zip' with types Array(String), {Int32, String, String, String}
Overloads are:
 - Hash(K, V)::zip(ary1 : Array(K), ary2 : Array(V))

        Hash.zip(field_names, row)
```

This PR solves the previous problem.

### DIFF
--- a/spec/pg/result_spec.cr
+++ b/spec/pg/result_spec.cr
@@ -48,6 +48,17 @@ describe PG::Result, "#to_hash" do
     ])
   end
 
+  it "represents the rows and fields as a hash with typed querying" do
+    res = DB.exec({String, String, Bool, Int32},
+                  "select 'a' as foo, 'b' as bar, true as baz, 10 as uhh
+                   union all
+                   select '', 'c', false, 20")
+    res.to_hash.should eq([
+      {"foo" => "a", "bar" => "b", "baz" => true,  "uhh" => 10},
+      {"foo" => "",  "bar" => "c", "baz" => false, "uhh" => 20}
+    ])
+  end
+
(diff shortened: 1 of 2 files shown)


will:
I'm not sure that the typed querying makes sense for to_hash. 

With this patch, 

```
    res = DB.exec({String, String, Bool, Int32},
                  "select 'a' as foo, 'b' as bar, true as baz, 10 as uhh
                   union all
                   select '', 'c', false, 20")
    p res.to_hash.first["uhh"] * 10
```

throws a compile-time error 

```
in ./spec/pg/result_spec.cr:60: undefined method '*' for Bool (compile-time type is (String | Int32 | Bool))

    p res.to_hash.first["uhh"] * 10
```

My original intent with the typed querying was to avoid union types, but since to_hash all ends up in one hash, you can't.

This should still be merged because to_hash shouldn't throw an error. Maybe though typed query exec should return an alternate kind of result that doesn't have to_hash.

will:
Thanks!

PR #22, opened 2015-10-07
Add escape_literal overload for binary data.

technorama:
Add spec for escape_literal.
Fix compiler/syntax error.

### DIFF
--- a/spec/pg/connection_spec.cr
+++ b/spec/pg/connection_spec.cr
@@ -85,6 +85,7 @@ end
 describe PG::Connection, "#escape_literal" do
   assert { DB.escape_literal(%(foo)).should eq(%('foo')) }
   assert { DB.escape_literal(%(some"thing)).should eq(%('some\"thing')) }
+  assert { DB.escape_literal(%(foo).to_slice).should eq(%('\\x666f6f')) }
   it "raises on invalid strings" do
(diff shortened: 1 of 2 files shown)


jhass:
Shouldn't it add an `E` in front? http://www.postgresql.org/docs/9.4/static/datatype-binary.html

technorama:
I don't believe so.  Escaped quoting is optional.

From the document you referenced:
> In some contexts, the initial backslash may need to be escaped by doubling it, in the same cases in which backslashes have to be doubled in escape format; 

> The reason multiple backslashes are required, as shown in Table 8-7, is that an input string written as a string literal must pass through two parse phases in the PostgreSQL server. The first backslash of each pair is interpreted as an escape character by the string-literal parser (assuming escape string syntax is used) and is therefore consumed, leaving the second backslash of the pair.

technorama:
@will Binary data is corrupted without this PR.  Crystal-pg doesn't even compile without the space before the colon.

will:
Thanks for the reminder. Interesting that the colon thing didn't break the specs though.

jhass:
It'll break with the next Crystal release, 0.8 is still fine with it.

technorama:
That must be why it didn't compile for me.  I'm using HEAD.

PR #25, opened 2015-10-20
Add connection with using Hash values.

tebakane:
Hi! 
I would like to add connection which using Hash values like `ruby-pg`.

    info = { host: "localhost", user: "postgres", password: "password", db_name: "test_db", port: "5432"}
    PG.connect(info)

### DIFF
--- a/spec/pg/connection_spec.cr
+++ b/spec/pg/connection_spec.cr
@@ -60,12 +60,11 @@ describe PG::Connection, "#exec untyped with params" do
   end
 
   it "can properly encode various types" do
-    time  = Time.new(2015,5,2,13,14,15,0, Time::Kind::Utc)
-    date  = Time.new(2015,5,2, 0, 0, 0,0, Time::Kind::Utc)
+    time = Time.new(2015, 5, 2, 13, 14, 15, 0, Time::Kind::Utc)
+    date = Time.new(2015, 5, 2, 0, 0, 0, 0, Time::Kind::Utc)
     slice = Slice(UInt8).new(UInt8[5, 0, 255, 128].to_unsafe, 4)
-    query = "select
-             $1::text, $2::int, $3::text, $4::float, $5::timestamptz, $6::date, $7::bool, $8::bytea"
-    param =  ["hello",       2,    nil,    -4.23,      time,            date,       true, slice]
+    query = "select $1::text, $2::int, $3::text, $4::float, $5::timestamptz, $6::date, $7::bool, $8::bytea"
+    param = ["hello", 2, nil, -4.23, time, date, true, slice]
     res = DB.exec(query, param)
     res.rows.should eq([param])
   end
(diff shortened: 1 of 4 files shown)


src/pg/connection.cr
```diff
@@ -37,6 +37,20 @@ module PG
...
+    # ```
+    # info = { host: "localhost", user: "postgres", password: "password", db_name: "test_db", port: "5432"}
+    # PG::Connection.new(info)
+    # ```
+    def initialize(conninfo : Hash)
+      info = ""
+      conninfo.each{|k,v|
+        info = info + k.to_s + "=" + v.to_s + " "
+      }
+      initialize(info)
```
jhass: How about `initialize(parameters.map {|param, value| "#{param}=#{value}" }.join(" "))`

src/pg/connection.cr
```diff
@@ -37,6 +37,20 @@ module PG
       end
     end
 
+    # `#initialize` Connect to the server with values of Hash.
+    #
+    # ```
+    # info = { host: "localhost", user: "postgres", password: "password", db_name: "test_db", port: "5432"}
+    # PG::Connection.new(info)
+    # ```
+    def initialize(conninfo : Hash)
```
jhass: `parameters : String|Symbol => String|Int` ?

tebakane: I think 
`parameters : YAML::Type => YAML::Type`
is better, because sometime connect with using yaml file.

src/pg/connection.cr
tebakane: ok! it is better than mine.

src/pg/connection.cr
```diff
@@ -37,6 +37,16 @@ module PG
       end
     end
 
+    # `#initialize` Connect to the server with values of Hash.
+    #
+    # ```
+    # info = { host: "localhost", user: "postgres", password: "password", db_name: "test_db", port: "5432"}
+    # PG::Connection.new(info)
+    # ```
+    def initialize(parameters : Hash(YAML::Type, YAML::Type))
```
will: Following on the discussion, are there any situations where the key/param side isn't just a `String`, so `Hash(String, YAML::Type)`

will:
Could you please add a test to `connection_spec.cr` that uses `URI.parse(DB_URL)` to attempt a new connection through this path. You may need to `require "uri"` in the spec

will:
Also thanks @jhass for the review :100:

src/pg/connection.cr
tebakane: if the parameters type is `Hash(String, YAML::Type)`,  error will happen.
error like `can't cast (Nil | String | Array(YAML::Type) | Hash(YAML::Type, YAML::Type)) to Hash(String, YAML::Type)`

    y = YAML.load("host: localhost\ndbname: postgres\nuser: postgres\npassword: postgres") as Hash
    PG.connect(y)

It is why I set the parameters type is `Hash(YAML::Type, YAML::Type)`.

PR #27, opened 2015-10-21
Add connection with using Hash values.

tebakane:
I'm sorry. I mistook and close PR.
https://github.com/will/crystal-pg/pull/25

Please discuss one more time about `Hash(YAML::Type, YAML::Type)` and `Hash(String, YAML:: Type)`.

Thanks.

### DIFF
--- a/spec/pg/connection_spec.cr
+++ b/spec/pg/connection_spec.cr
@@ -1,8 +1,17 @@
 require "../spec_helper"
-
+require "uri"
 describe PG::Connection, "#initialize" do
   it "works on a good connection" do
     PG::Connection.new(DB_URL)
+
+    uri = URI.parse(DB_URL)
+    info = {} of String => String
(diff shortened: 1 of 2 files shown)


will:
Thanks for adding the spec.

That makes sense if Crystal's YAML parser returns `YAML::Type`=>`YAML::Type`. However, I wonder though if we can instead go to just `initialize(parameters : Hash)` and let the compiler make sure that both sides of the hash have `#to_s` defined.

jhass:
That was what @tebakane tried initially and I was suggesting `String|Symbol => String|Int`. If we want to be open it should be just `Hash` indeed.

will:
Yeah, I see where that came from, but I think this works. I've merged this in and went ahead made that change. 

Thanks!

tebakane:
Thanks.

PR #30, opened 2015-12-16
Iterate rows and fields from results

ysbaddaden:
This allows to map a PG result to a struct or class, avoiding temporary memory structures (ie. Tuples and Hashes). I use this pattern extensively and would love to see it merged.

To achieve this, I keep a reference to the LibPQ::PGResult structure and delay the resolution of the rows and fields gatherer to be on demand. Actually only the rows have to, since fields is used by the iterator.

### DIFF
--- a/spec/pg/result_spec.cr
+++ b/spec/pg/result_spec.cr
@@ -64,3 +64,43 @@ describe PG::Result, "#to_hash" do
     expect_raises { res.to_hash }
   end
 end
+
+struct FooBarBaz
+  property foo, bar, baz
+
+  def initialize(@foo, @bar, @baz)
+  end
+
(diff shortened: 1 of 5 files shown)


will:
Just took a first look and this looks good, thanks. Going to go through it more later

Just idly thinking, don't feel obligated to do it, but do you think a macro would make sense to create the `.from_pg` and properties and initialize? Also if you know the order ahead of time, you could skip the case statement.

ysbaddaden:
In my use case, I don't know the order, but I guess the test could be simplified.

There could be a mapping macro. I'll see if I have some time to tacle it (not sure).

will:
Thanks

PR #31, opened 2015-12-17
Event loop

ysbaddaden:
I noticed that PG::Connection#exec was blocking the event loop. I followed http://www.postgresql.org/docs/9.3/static/libpq-async.html page, hacked a few things and... it actually worked.

Here is an example, which prints `done (i)` on the screen every second when the connection is evented. instead of waiting for `i` seconds between each prints:

```crystal
require "pg"

DB_URL = ENV["DATABASE_URL"]? || "postgres:///"

5.times do |i|
  spawn do
    db = PG.connect(DB_URL)
    db.exec("SELECT pg_sleep(#{i});")
    puts "done (#{i})"
  end
end

sleep
```

I pushed the pull request for feedback. It seems to be working fine, specs are passing, and the above test too. I verified we can execute multiple queries from the same coroutine (one after another) but executing queries to a single connection from different coroutines eventually raises an "another command is already in progress" error.

### DIFF
--- a/src/core_ext/scheduler.cr
+++ b/src/core_ext/scheduler.cr
@@ -1,0 +1,7 @@
+class Scheduler
+  def self.create_resume_event_on_read(fiber : Fiber, fd : Int)
+    @@eb.new_event(fd, LibEvent2::EventFlags::Read, fiber) do |s, flags, data|
+      (data as Fiber).resume
+    end
+  end
+end
--- a/src/pg/connection.cr
+++ b/src/pg/connection.cr
@@ -38,6 +38,10 @@ module PG
       end
     end
 
+    def finalize
+      finish
+    end
+
     # `#initialize` Connect to the server with values of Hash.
     #
     #     PG::Connection.new({ "host": "localhost", "user": "postgres",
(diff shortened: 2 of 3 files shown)


will:
Thanks for sending these in. I need a few days to decompress after travel before I can dig in, but I appreciate it.

Do you happen to know what's up with crystal-head failures on travis? It seems unrelated to the patches, but I also haven't looked into it yet

ysbaddaden:
Take your time!

I noticed a PCL error (resuming stalled coroutine), I guess the finalize method should do some cleanup.

The CI errors are related to JSON.parse returning a JSON::Any now (for simpler access of hash and arrays). I guess there could be a `==` delegate in that class.

ysbaddaden:
As expected I'm running into an issue with a pool of PG::Connection inside a HTTP::Server handler:

Regularly a fiber gets stuck forever at the `Scheduler.reschedule` line in `wait_readable`. I'm not exactly sure why. Maybe there should be a timeout on the socket event, and a manual call to `PQincomingInput`, at least to unstuck the fiber; I don't think it's possible for a race condition to happen between the event.add and the reschedule.

ysbaddaden:
I found the problem: don't memorize the read_event because it's tied to a fiber, and with a pool... you're likely to share the connection across fibers. Of course.

will:
Thanks the patch and tracking down the problems

ysbaddaden:
I'm starting to recover from a few sick days and you merged my pull requests, that will help me recover :-)

PR #34, opened 2016-03-31
Adds on_notice callback

radiospiel:


radiospiel:
@will I have no idea why the travis builds fail. I am running this locally with:

    DATABASE_URL=postgres://eno@localhost:5432/crystal-pg-test crystal spec

My local pg version is 9.4.1. My crystal is 0.14.2

will:
Thanks for the patch! I'm traveling now for conferences, so it'll be a bit before I can get to it, but I'll get to it as soon as I can.

will:
That travis failure is weird. I initially suspected a race, but not even running another select, or a sleep 1 make it pass, so maybe not.

Also you add 2 bindings from libpq but only use one, was that left in from an earlier attempt?

radiospiel:
Do the specs work on your machine - they do on mine. As for the additional binding: yes, was from a earlier attempt, but we might end up using the notice-receiver instead of the notice_processor. If I undertsnad everything right, libpg comes with a notice-receiver, which reads from the PGResult passed in, builds a message string, and then sends that to the active notice_processor. I guess this is your call to make :)

radiospiel:
Just had an idea: could it be that travis' postgres is set up to ignore notices below a certain level?

will:
That was it, I fixed it by setting the level to a known level, and merged. Thanks!

PR #36, opened 2016-04-14
Next crystal

will:
I'm having trouble with some of the changes for the new stuff in https://github.com/crystal-lang/crystal/pull/2443

I don't think `@read_event` is ever set in @ysbaddaden 's https://github.com/will/crystal-pg/pull/31 so commenting it out in https://github.com/will/crystal-pg/commit/1f52767eccca43fd3b4b14fd31bb8a41aa9eeae1 works

Other than that though, I've been stuck on a few cases with `crystal spec` and was wondering if @asterite or anyone could spare a few minutes. 

The two things so far:

- currently getting the types for `@rows` in result.cr
- before that the `fields[col].decoder.decode(val_ptr.to_slice(size))` line was complaining that a `fields[col]` was returning procs, but that isn't happening for me right now, also it doesn't return procs :/. That may have just been some other error I lost when I started over.

### DIFF
--- a/spec/pg/result_spec.cr
+++ b/spec/pg/result_spec.cr
@@ -66,7 +66,9 @@ describe PG::Result, "#to_hash" do
 end
 
 struct FooBarBaz
-  property foo, bar, baz
+  property foo : String?
+  property bar : Bool?
+  property baz : Int32?
 
   def initialize(@foo, @bar, @baz)
   end
--- a/src/pg/connection.cr
+++ b/src/pg/connection.cr
@@ -4,7 +4,7 @@ require "../core_ext/scheduler"
 module PG
   class Connection
     # :nodoc:
-    record Param, slice, format do # Internal wrapper to represent an encoded parameter
+    record Param, slice : Slice(UInt8), format : Int32 do # Internal wrapper to represent an encoded parameter
       delegate to_unsafe, slice
       delegate size, slice
 
--- a/src/pg/error.cr
+++ b/src/pg/error.cr
@@ -10,7 +10,7 @@ module PG
   end
 
   class ResultError < Error
-    def initialize(raw_result, @status)
+    def initialize(raw_result, @status : LibPQ::ExecStatusType)
       msg = String.new(LibPQ.result_error_message(raw_result))
       super(msg)
     end
--- a/src/pg/result.cr
+++ b/src/pg/result.cr
@@ -15,7 +15,7 @@ module PG
       property name
       property oid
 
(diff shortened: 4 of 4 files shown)


asterite:
For @rows it seems T is something like Tuple(Int32.class, String.class) and you need @rows to be Array(Tuple(Int32, String))... though I'm also seeing this:

```
     |  +- generic class PG::Result(Array(Array(JSON::Type) | Bool | Float32 | Float64 | Hash(String, JSON::Type) | Int32 | Int64 | String | Time | Nil)) (40 bytes)
     |         @types  : Array(Array(JSON::Type) | Bool | Float32 | Float64 | Hash(String, JSON::Type) | Int32 | Int64 | String | Time | Nil) (8 bytes)
     |         @res    : Pointer(Void)                                                                                                        (8 bytes)
     |         @fields : Array(PG::Result::Field)?                                                                                            (8 bytes)
     |         @rows   : Array(Array(Bool | Float32 | Float64 | Int16 | Int32 | Int64 | JSON::Any | Slice(UInt8) | String | Time | Nil))?     (8 bytes)
```

Here T is not a tuple and @rows is more or less T.

Right now there's no way to express that. But since `rows` will always be consumed, because I see very method in Result eventually calling `rows`, why not compute the rows and pass it to PG::Result, with T being the types of `rows`.

Then I think PG::Result::Row will maybe need to be generic too because you won't be able to just say `@result : Result` because it's generic.

The problem about the decoder is a bug, currently with this:

```crystal
class Foo(T)
  def initialize
    @array = Array.new(10, 0)
  end

  def array
    @array
  end
end
```

the compiler infers `@array` to be `Array`, without a generic type argument, and that's not currently possible so the error is garbage. I'll fix this in a few minutes.

asterite:
I'm also trying to think of a way to pass from a class type to its instance type, but I can't find it with the current language.

We have `T.class` in the type grammar to pass from instance to class. Maybe we are missing `T.instance` to do the inverse. With that you'll be able to do `@rows : Array(T.instance)`. Maybe a better name for that could be `T.new`, so you'd write `@rows : Array(T.new)`. What do you think?

will:
I managed to get all the specs passing. The row problem was fixed by removing the memoization.

I also had to make `Result::Row` into `Result::Row(T)`, but then it lost privileges to call the protected method

asterite:
Woops, that's a bug too. I'll fix it 😊

ysbaddaden:
Yes, please drop `@read_event` I didn't clean it up in my patch, sorry.

PR #37, opened 2016-05-01
Native driver

will:
This is incomplete, ~~notably no SSL or authentication other than `trust`~~ is supported yet. I'm opening this PR in hopes people run this on their own spec suites, to see if that uncovers any problem with the protocol implementation. This is mostly compatible with the released version, but there are a few incompatible changes.

native driver changes
=====================
new features
  * 100% crystal, using crystal's native async io (no longer using libpq)
  * each Connection#exec form can take a block, and stream the rows as they
      come in. This does not store any rows in memory and is very fast.

incompatable changes:
  * Result.each now yields the entire row and an array of fields
  * The notice callback now yields an entire
  * Error classes have changed


will:
I figured if Crystal 0.16 is released today, so should a native postgres driver

asterite:
By the way, do you have any performance benchmarks between the lib and native versions?

will:
Not yet, plan to do some benchmarking soon.

will:
I did one just myself, and it was comparable on a ~700MiB result set time wise, but with the streaming significantly less memory

will:
heh

```
crystal-pg ➤ ./native 
00:00:10.9976600
sum = 0
crystal-pg ➤ ./libpq   
Too many heap sections: Increase MAXHINCR or MAX_HEAP_SECTS
[1]    31669 abort      ./libpq
```

Issue #38, opened 2016-05-06
Exception when requesting `rows` of empty resultset


spalladino:
On version 0.7.0: Given an empty resultset, invoking `result.rows` yields `this should be unreachable`. For example:

```crystal
p db.exec("SELECT * FROM mytable where 1=0" ).rows
```

asterite:
The problem seems to be [this](https://github.com/will/crystal-pg/blob/master/src/pg/result.cr#L22):

```crystal
    def rows
      a = [first]
      each(@types) { |r| a << r }
      a[1..-1]
    end
```

I think this can be implemented like this:

```crystal
    def rows
      a = [] of typeof(first)
      each(@types) { |r| a << r }
      a
    end
```

@spalladino You can try redefining `Result#rows` in your project and see if that works.

spalladino:
That works, thanks @asterite! By the way, I was checking this with @waj, and he arrived at **exactly** the same code as you did.

will:
Thanks @spalladino for reporting, and @asterite for the fix :)

PR #39, opened 2016-05-06
Handle empty result sets

spalladino:
Fixes will/crystal-pg#38


will:
Sorry, race condition on fixing it. Thanks though!

spalladino:
Better having more than one fix than having none! Thanks for the quick response, will!

PR #45, opened 2016-05-29
Add missing `true` in PQ::Connection#handle_async_frames

asterite:
Adds a missing `true`. It works without it too, but it's just coincidence. `handle_notification` returns `Void`, and right now it's considered truthy. I plan to make some changes to `Void` and without this fix this code breaks.


will:
Thanks!

PR #47, opened 2016-06-14
(incomplete) compile on 0.18.0

will:
Still openssl spec failures though

### DIFF
--- a/src/pq/connection.cr
+++ b/src/pq/connection.cr
@@ -3,7 +3,7 @@ require "crypto/md5"
 require "socket"
 require "socket/tcp_socket"
 require "socket/unix_socket"
-require "openssl/openssl"
+require "openssl"
 
 # DEBUG = ENV["DEBUG"]?
 
@@ -12,7 +12,7 @@ module PQ
 
   # :nodoc:
   class Connection
-    getter soc : UNIXSocket | TCPSocket | OpenSSL::SSL::Socket
+    getter soc : UNIXSocket | TCPSocket | OpenSSL::SSL::Socket::Client
     getter server_parameters : Hash(String, String)
     property notice_handler : Notice ->
     property notification_handler : Notification ->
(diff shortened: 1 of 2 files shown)


jhass:
Mmh, which specs?

will:
There is a file you need to touch to enable some of the auth specs. They're
not on by default because they need some setup on your local Postgres
server that isn't standard with brew/apt/etc

jhass:
Yup found that, they pass for me.

will:
`SSL_connect: error:14090086:SSL routines:SSL3_GET_SERVER_CERTIFICATE:certificate verify failed (OpenSSL::SSL::Error)`

I'm just using a self-signed cert, I think the default verification level went up in this release.

jhass:
Yes it did, I guess it works for me because I didn't explicitly configure my postgres to enable TLS. You should get a way with abusing your self signed cert as a CA certificate too:
```cr
context = OpenSSL::SSL::Context::Client.new
context.ca_certificates = ENV["POSTGRESQL_SERVER_CERTIFICATE"]
socket  = OpenSSL::SSL:Socket::Client.new(socket, context: context, hostname: host)
```
Of course proper setups should either use their own proper CA cert here and create their PostgreSQL server certificate from it, or use some blessed PKIX.

will:
libpq supports a few different levels https://www.postgresql.org/docs/current/static/libpq-ssl.html table 31-1

jhass:
* `disable`: Don't create a `OpenSSL::SSL:Socket::Client` at all. Probably refuse connection if the server demands it.
* `allow`: Application protocol specific, don't request from the client, but create a `OpenSSL::SSL::Socket::Client` in `require` mode.
* `prefer`: Application protocol specific, request from the client, create a `OpenSSL::SSL::Socket::Client` in `require` mode if the server supports it.
* `require`, create a `OpenSSL::SSL::Socket::Client`, pass a context with `verify_mode` set to `OpenSSL::SSL::VerifyMode::NONE`, the `hostname` parameter of the socket should get ignored.
* `verify-ca`: create a standard `OpenSSL::SSL::Socket::Client` with no additional options, do not pass the `hostname` parameter.
* `verify-full`: create a standard `OpenSSL::SSL::Socket::Client` with the `hostname` parameter.

That is if you would want to mirror what libpq does, I'm not sure I would recommend that. Allowing the client to pass a `OpenSSL::SSL::Context::Client` might be easier and a bit more flexible, it's what `HTTP::Client` does.

Issue #48, opened 2016-06-15
Use types instead of tuples, arrays of tuples, etc.?


asterite:
Yesterday we tried to write some code with @ggiraldez that queried postgres to get some GIS data. It took us a while to figure out what types were returned by the driver, or what types we needed to pass to `execute`. I can't remember what we ended up passing, I think `Tuple` and then casting each member to `Float64`, with `as`.

Given that Crystal allows user-defined types, I think it would be much more convenient and intuitive if, instead of returning tuples and arrays of tuples, proper types would be returned. They can be structs to preserve performance. Basically something like this:

```crystal
record PG::Geo::Point, x : Float64, y : Float64

# ...

class PointDecoder < Decoder
  def decode(bytes)
    x = swap64(bytes)
    y = swap64(bytes + 8)

    Point.new(
      (pointerof(x).as(Float64*)).value,
      (pointerof(y).as(Float64*)).value,
    )
  end
end
```

For `PolygonDecoder` with can return an `Array(Point)`, and maybe have an `alias Polygon = Array(Point)`, for `PathDecoder` we'd return a `Geo::Path` struct that has an open/closed Bool property, along with the points, etc.

That way one can specify `Point` in the query instead of `Tuple(Float64, Float64)`. This is both less code to read/write, more clear, and has the same performance.

There's another issue: tupe types of the same size get merged. So `BoxDecoder` returns a two-elements tuple, and so does `PointDecoder`, so one can't specify `Tuple(Float64, Float64)` as a target type because the real type returned by all decoders is a tuple where each component is a union.

I can create a PR with code for this, but I wanted to know first if you like this idea.

will:
Yeah I think that makes a lot sense.

Also while you're in there, I've been planning on removing the opt-into geo decoders part; I don't think it saves much.

asterite:
Great! I'll send a PR in a few days :-)

PR #49, opened 2016-06-16
Add property Geo types. Implements #48

asterite:
I decided to add types for things whose representation if the same, for example Box and LIneSegment, mostly because they represent different things, and one could imagine adding different methods to them. The same applies to Line and Circle. For Path I couldn't use `record` because I needed a query method (`closed?`), I'll think how we can make `record` do that automatically.


will:
Thank you!

PR #55, opened 2016-08-12
segfault maybe in spec framework?

will:
Very strange thing here @asterite ,  it looks like it might not be in the my code but in the spec framework. I'm nowhere near sure though, I could be very mistaken.

Printing the hexdump with a particular offset triggers the segfault. Any other offset, and the segfault doesn't happen. But the segfault is not happening during the hexdump, as other prints happen. It looks like it's happening when gathering the `#inspect` for the error messages. 

```
F
[36, 16]
[1, false, 23, [{dim: 2, lbound: 1}]]
Slice[0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 23, 0, 0, 0, 2, 0, 0, 0, 1, 0, 0, 0, 4, 0, 0, 0, 1, 0, 0, 0, 4, 0, 0, 0, 2]
0000 0004 0000 0001 0000 0004 0000 0002  ................
Slice[0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 23, 0, 0, 0, 2, 0, 0, 0, 1, 0, 0, 0, 4, 0, 0, 0, 1, 0, 0, 0, 4, 0, 0, 0, 2]
ok
Invalid memory access (signal 11) at address 0x0
[4337667195] *CallStack::print_backtrace:Int32 +107
[4337596263] __crystal_sigfault_handler +55
[140735510512938] _sigtramp +26
[4339653361] GC_generic_malloc_many +560
[4339681858] GC_malloc_atomic +132
[4337579899] __crystal_malloc_atomic +11
[4337719657] *GC::malloc_atomic<UInt32>:Pointer(Void) +9
[4337682213] *String::Builder#initialize<Int32>:Bool +37
[4337682154] *String::Builder::new<Int32>:String::Builder +138
[4337678625] *Array(Int32)@Object#inspect:String +33
[4338464745] *Spec::EqualExpectation(Array(Int32))@Spec::EqualExpectation(T)#failure_message<Array(Int32)>:String +41
[4337680255] *Array(Int32)@Spec::ObjectExtensions#should<Spec::EqualExpectation(Array(Int32)), String, Int32>:Nil +63
[4337644720] ~proc10Proc(Nil)@./spec/pg/decoder_spec.cr:4 +400
[4337625895] *it<String, String, Int32, &Proc(Nil)>:(Array(Spec::Result) | Nil) +391
[4338260175] */Users/will/code/crystal-pg/spec/pg/decoder_spec.cr#test_decode<String, String, Array(Int32), String, Int32>:(Array(Spec::Result) | Nil) +111
[4337624424] ~procProc(Nil)@./spec/pg/decoder_spec.cr:12 +3272
[4337918477] *Spec::RootContext::describe<String, String, Int32, &Proc(Nil)>:Spec::Context+ +365
[4337621091] *describe<PG::Decoders:Module, String, Int32, &Proc(Nil)>:Spec::Context+ +51
[4337540998] __crystal_main +4374
[4337595912] main +40
```

The `[4315101473] *Array(Int32)@Object#inspect:String +33` line changes type with whatever the return of `decode` is. For example changing the return to a string array shows `*Array(String)@Object#inspect` instead.

Sorry I don't have a better handle on this before grabbing your attention.

### DIFF
--- a/spec/pg/decoder_spec.cr
+++ b/spec/pg/decoder_spec.cr
@@ -67,6 +67,15 @@ describe PG::Decoders do
     x.call("nan").nan?.should be_true
   end
 
+  test_decode "array", "ARRAY[1]", [1]
+  test_decode "array", "ARRAY[1,2]", [1,2]
+  test_decode "array", "ARRAY[ARRAY[9,8],ARRAY[7,6]] ", [[1,2],[3,4]]
+  test_decode "array", "'{{{1,2},{3,3}},{{2,5},{4,3}}}'::integer[]", [[1,2],[3,4]]
+  exit
+  test_decode "array", "ARRAY[1, null] ", [1, nil]
+  test_decode "array", "('[10:12]={1,2,3}'::integer[])", 10
+#  test_decode "int array ", "ARRAY[1,2,3,4,5,6,7]", [1, 2, 3, 4, 5, 6, 7]
+
   test_decode "xml", "'<json>false</json>'::xml", "<json>false</json>"
   test_decode "char", %('c'::"char"), 'c'
   test_decode "bpchar", %('c'::char), "c"
--- a/src/pg/decoder.cr
+++ b/src/pg/decoder.cr
@@ -111,6 +111,57 @@ module PG
       end
     end
 
+    alias PGInt32Array = Int32? | Array(PGInt32Array)
+
+    class ArrayDecoder < Decoder
+      def decode(bytes)
+        puts
+        dimensions = swap32(bytes).to_i
+        has_null = swap32(bytes + 4) == 1 ? true : false
+        oid = swap32(bytes + 8)
+        dim_info = Array(NamedTuple(dim: UInt32, lbound: UInt32)).new(dimensions) do |i|
+          offset = 12 + (8*i)
+          {
+            dim: swap32(bytes + offset),
+            lbound: swap32(bytes + (offset + 4))
+          }
+        end
+        data_start = (8*dimensions)+8+4
+        puts [bytes.size, (bytes+data_start).size]
+
+
+        p [dimensions, has_null, oid, dim_info]
+        p bytes
+
+        # using any other offset here, such as +1 or -1 avoids the segfualt
+        puts (bytes + (data_start )).hexdump
+
+        #but the hexdump itself is not segfualting, these next two prints make it
+        p bytes
+        puts "ok"
+
+
+        #Array(PGInt32Array).new(dim_info[0][:dim].to_i) do |i|
+        #  get_element(bytes, dim_info, 0, i)
+        #end
+
+        # the segfault includes the type of the result here
+        # ["hi"]  # [4344595745] *Array(String)@Object#inspect:String +33
+        # vs
+        [3]  # [4315101473] *Array(Int32)@Object#inspect:String +33
+      end
+
+      def get_element(bytes, dim_info, depth, pos)
+        if depth == (dim_info.size-1)
+          0
+        else
+          1
+        end
+      end
+
+
+    end
+
     class PointDecoder < Decoder
       def decode(bytes)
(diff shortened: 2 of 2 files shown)


asterite:
@will Thank you, these issues are always very important to fix, and also fun for me :-)

I checked out the `segfault` branch and did `crystal spec`, but didn't get any segfault. What should I do to trigger it?

asterite:
Also, whenever you get a segfault one quick test one can do is to run again the program with `GC_DONT_GC=1`. If it doesn't segfault it means it has something to do with the GC, probably a reference that is being cleared by the GC but shouldn't.

will:
@asterite try running just `crystal spec spec/pg/decoder_spec.cr`. I just noticed when running the full suite `crystal spec` there is no segfault.

The `GC_DONT_GC` env var doesn't seem to have any change.

asterite:
@will Seems to be a bug in the old `Slice#hexdump`

Try using this implementation, the commented line is the only change I made to the old implementation. You'll see it raises IndexError, so it's writing something out of place. It doesn't happen with master's hexdump, but I'll review it now to add these safety checks just in case (speed is not very important in hexdump).

```cr
struct Slice
  def hexdump
    self.as(Slice(UInt8))

    full_lines, leftover = size.divmod(16)
    if leftover == 0
      str_size = full_lines*58 - 1
    else
      str_size = (full_lines + 1)*58 - (16 - leftover) - 1
    end

    String.new(str_size) do |buf|
      hex_offset = 0
      ascii_offset = 41

      # Detect index out of bounds
      buffer = Slice.new(buf, str_size)

      each_with_index do |v, i|
        buffer[hex_offset] = to_hex(v >> 4)
        buffer[hex_offset + 1] = to_hex(v & 0x0f)
        hex_offset += 2

        buffer[ascii_offset] = (v > 31 && v < 127) ? v : '.'.ord.to_u8
        ascii_offset += 1

        if i % 2 == 1
          buffer[hex_offset] = ' '.ord.to_u8
          hex_offset += 1
        end

        if i % 16 == 15
          buffer[hex_offset] = ' '.ord.to_u8
          buffer[ascii_offset] = '\n'.ord.to_u8
          ascii_offset += 42
          hex_offset += 18
        end
      end

      while hex_offset % 58 < 41
        buffer[hex_offset] = ' '.ord.to_u8
        hex_offset += 1
      end

      {str_size, str_size}
    end
  end
end
```

asterite:
Same error in master... I'll fix it.

asterite:
For reference, it happens with a slice of exactly 16 elements

asterite:
https://github.com/crystal-lang/crystal/commit/f1b05053d63fe583974b6376b21da08ebe5556cd

asterite:
@will @ysbaddaden Would be nice if you could check the code to see why there was an index out of bounds. My guess is that that line was to append a newline, but we don't need to do that if there's no next line... but I'm not sure. In any case now it's protected by a slice access so at most we get an exception, not a random segfault.

will:
Just curious but do you know why it segfaults after the hexdump is done and
also after other stdout, and then also only sometimes?

asterite:
@will Writing outside allocated memory is usually unpredictable (at least in my experience, I honestly never did much of C/low-level coding, I'd say Crystal is the first project where I really came close to the metal). I think you can't reliably know what are you writing to. In this case heap memory was overwritten, so it was probably overwriting some other object in heap memory. Since what's in heap memory depends on what you do (what calls create heap objects), that could explain why it happens in some cases.

will:
@asterite 
> My guess is that that line was to append a newline, but we don't need to do that if there's no next line

This is exactly the case. While working on it, I did check exact matches of 16, but since, as you pointed out, the heap overwrite can have random effects, never came across a segfault in testing, so didn't notice the problem until this. 

The patch looks good, thanks for finding the problem.

PR #60, opened 2016-10-03
crystal-db

asterite:
This makes crystal-pg compatible with [crystal-db](https://github.com/crystal-lang/crystal-db)

I'm marking this as WIP only because there might be some leftovers from the old code, and I could rewrite commit history for better messages, but mostly to open the discussion.

If you want to get a taste of how to use PG now, here's some sample code:

<details>
  <title>Sample code</title>

``` cr
  require "./src/pg"

  # Don't worry: this isn't always used in the examples below
  class Contact
    DB.mapping(name: String, age: Int32)
  end

  DB.open "postgres:///" do |db|
    db.exec "drop table if exists contacts"
    db.exec "create table contacts (name varchar, age int4)"
    db.exec "insert into contacts values ($1, $2)", "John Doe", 30

    # `read(*types)` to read multiple values at once
    db.query "select name, age from contacts" do |rs|
      rs.each do
        name, age = rs.read(String, Int32)
        pp name, age
        # name # => "John Doe"
        # age  # => 30
      end
    end

    # `query_one` to read a single row
    # This raises if there are no rows, or more than one row
    name, age = db.query_one "select name, age from contacts", &.read(String, Int32)
    pp name, age
    # name # => "John Doe"
    # age  # => 30

    # Another way:
    name, age = db.query_one "select name, age from contacts", as: {String, Int32}
    pp name, age
    # name # => "John Doe"
    # age  # => 30

    # Another way, just read the name:
    name = db.query_one "select name from contacts", as: String
    pp name
    # name # => "John Doe"

    # similar to `query_one`, but returns nil if there are no rows (still raises if there are many rows)
    result = db.query_one? "select name, age from contacts", &.read(String, Int32)
    pp result, typeof(result)
    # result         # => {"John Doe", 30}
    # typeof(result) # => (Tuple(String, Int32) | Nil)

    # Another way:
    result = db.query_one? "select name, age from contacts", as: {String, Int32}
    pp result, typeof(result)
    # result         # => {"John Doe", 30}
    # typeof(result) # => (Tuple(String, Int32) | Nil)

    # Another way, just read the name:
    name = db.query_one? "select name from contacts", as: String
    pp name, typeof(name)
    # name         # => "John Doe"
    # typeof(name) # => String | Nil

    # Another way, use DB.mapping
    contact = db.query_one "select name, age from contacts", as: Contact
    p contact

    db.exec "insert into contacts values ($1, $2)", "Foo Bar", 40

    # `query_all` to read an array of values
    people = db.query_all "select name, age from contacts order by age desc", &.read(String, Int32)
    pp people, typeof(people)
    # people         # => [{"Foo Bar", 40}, {"John Doe", 30}]
    # typeof(people) # => Array(Tuple(String, Int32))

    # Another way:
    people = db.query_all "select name, age from contacts order by age desc", as: {String, Int32}
    pp people, typeof(people)
    # people         # => [{"Foo Bar", 40}, {"John Doe", 30}]
    # typeof(people) # => Array(Tuple(String, Int32))

    # Another way, just read the name:
    names = db.query_all "select name from contacts order by age desc", as: String
    pp names, typeof(names)
    # names         # => ["Foo Bar", "John Doe"
    # typeof(names) # => Array(String)

    # Another way, use DB.mapping
    contacts = db.query_all "select name, age from contacts", as: Contact
    p contacts

    # Reading geo data is easy too
    point = db.query_one "select '(1.2,3.4)'::point", as: PG::Geo::Point
    p point

    # Or an array, for that matter
    array = db.query_one "select '{{9,8},{7,6},{5,4}}'::integer[]", as: Array(Array(Int32))
    p array
  end
```

</details>

All types that were previously supported are still supported, including Geo types and arrays. Crystal DB provides an argless `read` method: this returns a union of all possible types supported by the DB. When reading arrays, you'll get these recursive arrays for which decoders are registered. However, you can read a column and specify a target type: in that case you read a specific array type, for example by doing `rs.read(Array(Array(Int32))`, which should make coding against these types much easier (this was possible in the old implementation and I thought it would be nice to keep it).

To use methods not exposed by crystal-db's API you can do:

``` cr
db = DB.open("postgres:///")
db.using_connection do |conn|
  # Cast isn't necessary if this is the only registered driver
  conn = conn.as(PG::Connection)
  # Now invoke whatever methods you need, for example on_notice
end
```

Results are streamed, like before, and in fact for big data it should be a bit faster because data is decoded directly from the socket: previously a whole slice was read first and then decoded. I didn't do any benchmarks to compare the codes, though, because I don't know what would be a good benchmark.

One small annoying thing is that we can't use a constant `DB` to hold the database connection anymore. I used `PG_DB`. Maybe it would be better to just rename `DB` to something else, because it'll only be used for `DB.open`... but that's another topic :-)


asterite:
Oh, I forgot something: I'm missing a way to get the number of affected rows. Is this possible? I know getting the last id isn't possible, but maybe affected rows is...?

will:
I'm out right now on mobile, but you can get the Id with like insert ...
returning id

I think there might be a protocol frame I'm ignoring right now that has
affected rows, but it might be a string saying "7 rows affected" like in
psql. Need to double check about that though

asterite:
Found the affected rows info: it comes in a CommandComplete frame as a string.

I also added support for encoding (insert, update) arrays and geo types, something that was not supported before this PR :-)

asterite:
No idea why specs are failing :-(

(it's also strange that the jsonb specs are running on travis that don't have pg >= 9.4)

spec/pq/authentication_methods_spec.cr
```diff
@@ -9,38 +9,65 @@ require "../spec_helper"
...
-      PG::Connection.new("host=localhost user=crystal_test")
+      DB.open("postgres://crystal_test@localhost")
     }
-    DB.exec("drop role if exists crystal_test")
+    PG_DB.exec("drop role if exists crystal_test")
   end
 end
 
 if File.exists?(File.join(File.dirname(__FILE__), "../.run_auth_specs"))
+  describe PQ::Connection, "cleartext auth" do
```
will: did the cleartext auth get readded from your branch? I removed it previously to prevent a mitm attack, and also there aren't really any servers left that don't support the md5 auth.

src/pg.cr
```diff
@@ -1,13 +1,18 @@
...
 
   # Establish a special listen connection to the database
-  def self.connect_listen(conninfo, *channels : String, &blk : PQ::Notification ->)
-    ListenConnection.new(conninfo, *channels, &blk)
+  def self.connect_listen(url, *channels : String, &blk : PQ::Notification ->)
+    db = DB.open(url)
+    db.using_connection do |conn|
+      conn.as(PG::Connection).listen(*channels, &blk)
+    end
+    db
```
will: The reason before that I returned a different class here was to make sure you can't accidentally try and use it to send normal queries on it, because when the driver is in the listen loop it wont work.

src/pg/connection.cr
```diff
@@ -1,148 +1,56 @@
...
 
-    # Connect with a `PQ::ConnInfo`.
-    #
-    # If you are making a lot of connections to the same database, it may be
-    # useful to reuse the `PQ::ConnInfo` to avoid reparsing the connection
-    # every time. All other initialization methods end up here.
-    def initialize(conninfo : PQ::ConnInfo)
-      @pq_conn = PQ::Connection.new(conninfo)
-      @pq_conn.connect
-      exec("SET extra_float_digits = 3")
```
will: I'm not sure if extra_float_digits is still necessary, so leaving this comment until I get confirmation.

src/pq/param.cr
```diff
@@ -7,22 +7,93 @@ module PQ
 
     # The only special case is nil->null and slice.
     # If more types need special cases, there should be an encoder
```
will: hah, should probably kill this comment

spec/pq/authentication_methods_spec.cr
asterite: Oops, seems I readded it. I'll remove it.

src/pg.cr
asterite: I'll change it back to return a ListenConnection.

asterite:
OK, it's failing now for me too after I upgrade crystal-db... this will be... fun :-P

src/pg/connection.cr
will: I'm pretty sure that this isn't necessary anymore. I think think it's just for the text result format, and after switching to binary result format it doesn't do anything. Like 95% sure.

uhoh-itsmaciek:
@will @asterite there kind of is something at the [protocol level](https://www.postgresql.org/docs/current/static/protocol-message-formats.html):

>CommandComplete (B)
>Byte1('C')
>Identifies the message as a command-completed response.
>
>Int32
>Length of message contents in bytes, including self.
>
>String
>The command tag. This is usually a single word that identifies which SQL command was completed.
>
>For an INSERT command, the tag is INSERT oid rows, where rows is the number of rows inserted. >oid is the object ID of the inserted row if rows is 1 and the target table has OIDs; otherwise oid is 0.
>
>For a DELETE command, the tag is DELETE rows where rows is the number of rows deleted.
>
>For an UPDATE command, the tag is UPDATE rows where rows is the number of rows updated.
>
>For a SELECT or CREATE TABLE AS command, the tag is SELECT rows where rows is the number >of rows retrieved.
>
>For a MOVE command, the tag is MOVE rows where rows is the number of rows the cursor's position >has been changed by.
>
>For a FETCH command, the tag is FETCH rows where rows is the number of rows that have been >retrieved from the cursor.
>
>For a COPY command, the tag is COPY rows where rows is the number of rows copied. (Note: the >row count appears only in PostgreSQL 8.2 and later.)
```

So if you match the command complete "tag" using something like `/.+ (\d+)$/` you can get a number that is most likely the number of rows affected, if there is a match. Note that the list above is not exhaustive: e.g., DDL commands have their own tags, and other assorted commands do to. E.g., `DO`, `LISTEN`, `NOTIFY`--I *think* the tag is what is echoed after you issue a command in psql. But if there is a number there, I haven't seen it be anything other than affected rows.

You can get the id in the result set with `RETURNING`, but right, there's no way to get the id without modifying the query (the INSERT will give you the oid on a `WITH OIDS` table, but that's basically never useful).

**Comment by uhoh-itsmaciek** [comment]
Sorry, had a stale version of this opened apparently--you found the relevant info.

File: src/pg/connection.cr (hunk 3 above)
Comment (uhoh-itsmaciek) [inline src/pg/connection.cr]: Yeah, `extra_float_digits` only matters for the text encoding. If you're doing everything in binary you can skip it.

**Comment by RX14** [comment]
What's left to be done in this PR?

**Comment by asterite** [comment]
@RX14  Nothing, I think it's ready to go

**Comment by will** [comment]
Oh, I never got emailed when the new patches came in just the comments, so I assumed it was still in "this will be… fun" mode.

I have really sketchy availability this week from thanksgiving, and next week for the aws conference, but I'll start taking a look at this.

**Comment by asterite** [comment]
@will Take your time, enjoy thanksgiving and the conference :-)

**Comment by will** [comment]
@asterite was there a change in 0.20 that would cause this?

```
crystal-pg ➤ crystal version                                                                                                                                                            git:505cfac*
Crystal 0.20.1 (2016-12-05)
crystal-pg ➤ crystal deps                                                                                                                                                               git:505cfac*
Updating https://github.com/crystal-lang/crystal-db.git
Using db (cd10dabba26a3d138a7fe12cbce110a421cc5934)
crystal-pg ➤ crystal spec                                                                                                                                                               git:505cfac*
Error in line 1: while requiring "./spec/pg/connection_spec.cr"

in spec/pg/connection_spec.cr:43: instantiating 'DB::Database#exec(String)'

    PG_DB.exec("listen somechannel")
          ^~~~

in lib/db/src/db/query_methods.cr:203: instantiating 'DB::PoolStatement#exec(Tuple())'

      prepare(query).exec(*args)
                     ^~~~

in lib/db/src/db/pool_statement.cr:33: instantiating 'statement_with_retry()'

      statement_with_retry &.exec
      ^~~~~~~~~~~~~~~~~~~~

in lib/db/src/db/pool_statement.cr:81: instantiating 'DB::Database#retry()'

      @db.retry do
          ^~~~~

in lib/db/src/db/database.cr:90: instantiating 'DB::Pool(DB::Connection)#retry()'

      @pool.retry do
            ^~~~~

in lib/db/src/db/database.cr:90: instantiating 'DB::Pool(DB::Connection)#retry()'

      @pool.retry do
            ^~~~~

in lib/db/src/db/pool_statement.cr:81: instantiating 'DB::Database#retry()'

      @db.retry do
          ^~~~~

in lib/db/src/db/pool_statement.cr:33: instantiating 'statement_with_retry()'

      statement_with_retry &.exec
      ^~~~~~~~~~~~~~~~~~~~

in lib/db/src/db/pool_statement.cr:33: instantiating 'DB::Statement+#exec()'

      statement_with_retry &.exec
                             ^~~~

in lib/db/src/db/statement.cr:67: can only use primitive integers and floats with Slice.new(size), not (Float32 | Float64 | Int32 | Int64 | Slice(UInt8) | String | Nil)

      perform_exec_and_release(Slice(Any).new(0))
                                          ^~~


```

**Comment by asterite** [comment]
Yes, it's this one: https://github.com/crystal-lang/crystal/issues/3624

I pushed a fix to crystal-db some minutes ago :-)

BTW, although that looks like a regular compiler error, it's actually a regular `raise` call in a macro: https://github.com/crystal-lang/crystal/blob/284f7d6bdf963dfd48fef2d70e5ed20d71b39815/src/slice.cr#L78-L80 . Now the compiler will show these better, inlined, as showing the location of the `raise` call didn't add much information (in fact, it made the error harder to understand).

File: shard.yml [hunk 5]
```diff
@@ -1,2 +1,7 @@
 name: pg
 version: 0.11.0
+
+dependencies:
+  db:
+    github: crystal-lang/crystal-db
+    branch: master
```
bcardiff: @asterite it should be `version: ~> 0.2.2`. So crystal-db/master can move forward without breaking will/crystal-pg master.

Or the releases of crystal-pg should be on a branch where the db version is fixed, but that is too much burden.

As soon as we crystal-pg onto db I will move forward with [unprepared statements](https://github.com/crystal-lang/crystal-db/pull/25) / [transactions](https://github.com/crystal-lang/crystal-db/tree/feature/transactions) on all 3 drivers :-)

asterite: Fixed!

spec/pq/authentication_methods_spec.cr
```diff
@@ -9,38 +9,39 @@ require "../spec_helper"
...
-      DB.exec("drop role if exists crystal_md5")
-      DB.exec("create role crystal_md5 login encrypted password 'pass'")
-      conn = PG::Connection.new("host=localhost user=crystal_md5 password=pass")
-      conn.exec("select 1").rows.first.first.should eq(1)
-      DB.exec("drop role if exists crystal_md5")
+      PG_DB.exec("drop role if exists crystal_md5")
+      PG_DB.exec("create role crystal_md5 login encrypted password 'pass'")
+      DB.open("postgres://crystal_md5:pass@localhost") do |db|
+        db.query_one("select 1", &.read).should eq(1)
+      end
```
asterite: How is this accomplished in golang or java? As far as I know, both languages use URI connection strings too, so if they have a method of doing that, we should use it too.

bcardiff: For the record, i think the github way to patch a PR-1 is send a PR-2 based on the branch the PR-1 is based on.

For the length of the patch your post is clearly lighter ;-)

sdogruyol:
This look great 💯 Is there any ETA for this since a lot of people (including me) waiting for PG support 😄

tbrand:
Including me as well 😄

asterite:
@will I applied the patch :-)

RX14:
:tada:

will:
Released v0.12.0 with this. Still need to update examples and instructions in the readme, but this looks good. Thanks @asterite :)

PR #62, opened 2016-10-28
Added macro so you can use either query or exec

puppetpies:
Hi Will,

I am not sure if you want to merge but other db driver usually have a query function for executing a query however yours has exec so i made a simple macro.

My reason is i want to use a same variable for multiple database drivers which all have query so for the union type to work i need this.

Otherwise i just use my own fork ?

### DIFF
--- a/src/pg/connection.cr
+++ b/src/pg/connection.cr
@@ -50,64 +50,66 @@ module PG
     def listen
       spawn { @pq_conn.read_async_frame_loop }
     end
-
-    # Execute an untyped query and store the results in memory.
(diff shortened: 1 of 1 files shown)


asterite:
@puppetpies If we ever merge #60 then all APIs will be the same across drivers

will:
Yes, the plan is to go with #60. In the meantime you could probably reopen
the class in your own code and use alias. No need to fork

will:
#60 is in so I'm going to close this. Thanks for the submission though!

PR #66, opened 2016-12-13
update to db unprepared statements

bcardiff:
From https://github.com/crystal-lang/crystal-db/pull/25 (currently in master but not released) a connection will create prepared and non prepared statements. This is required for uniform transactions api.

It seems crystal-pg has only implemented non prepared statements so the quick solution is to reuse them when ever a prepared statement is expected.

Prepared statements have special treatment since they need the be released as opposed to non prepared ones.

I would vouch for a proper prepared statement in postgres, but I am not sure you will agree.

### DIFF
--- a/src/pg/connection.cr
+++ b/src/pg/connection.cr
@@ -11,7 +11,11 @@ module PG
       @connection.connect
     end
 
-    def build_statement(query)
+    def build_prepared_statement(query)
+      Statement.new(self, query)
+    end
+
+    def build_unprepared_statement(query)
       Statement.new(self, query)
     end
 


samueleaton:
@bcardiff @will This is breaking stuff.

```
Error in libs/db/db/connection.cr:36: abstract `def DB::Connection#build_statement(query)` must be implemented by PG::Connection

    abstract def build_statement(query) : Statement
                 ^~~~~~~~~~~~~~~
```

bcardiff:
@samueleaton yes but only if you depend on crystal-pg master branch for your app/lib. 

The alternative is that in master of this shard we keep crystal-db master as dependency and upon publishing a.new version of -pg that is fixed to some versioning number. Or a full git flow scheme is followed. 

In the following days a new release of db will come and master will be stable again.

samueleaton:
Should it be in a different working branch until it works? Seems premature to put it in master.

But yeah I removed my reference to master. 👍

will:
> Seems premature to put it in master.

I misunderstood and thought not having this merged was blocking finishing the other patch

bcardiff:
Apologies I should been more asserting when I said
> (currently in master but not released)

At the end of the day it depends if we see master as stable or unstable.
Also, the story of developing multiple shards at the same time [is not nice enough still](https://github.com/crystal-lang/shards/issues/132).

I've just sent another PR that will fix master since crystal-db 0.3.0 is released now.

PR #71, opened 2016-12-15
Enable connection pool

bcardiff:
This PR enables the connection pooling from crystal-db to work with crystal-pg.

In order for the connection pool to work `DB::ConnectionLost` and `DB::ConnectionRefused` need to be raised when executing commands or establishing connections.

This also requires to crystal-db ~> ~~0.3.1~~ 0.3.2 

the fix for `pq/authentication_methods_spec.cr` could be different, those specs uses `/pg/*` so now the exception to expect need to be `DB::ConnectionLost` and `DB::ConnectionRefused`.

cc: @raydf

### DIFF
--- a/shard.yml
+++ b/shard.yml
@@ -4,4 +4,4 @@ version: 0.12.0
 dependencies:
   db:
     github: crystal-lang/crystal-db
-    version: ~> 0.3.0
+    version: ~> 0.3.1
--- a/shard.yml
+++ b/shard.yml
@@ -4,4 +4,4 @@ version: 0.12.0
 dependencies:
   db:
     github: crystal-lang/crystal-db
-    version: ~> 0.3.1
+    version: ~> 0.3.2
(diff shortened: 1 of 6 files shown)


bcardiff:
I've tested the pool retry logic with the following script.

https://gist.github.com/bcardiff/0cef66c4316ff07ba5b6f1c5c1e827b3

the script will perform queries and every now and then kill the connections. It also tolerate the server to go down for about 24sec (8 retries with 3 sec wait) without forwarding the reconnection to the user.

the require can be changed to `require "../src/pg"` in order to use your current working copy.

Nothing pending to be done from my side, I added the [WIP] because I wanted to do some more testing with that script. I think is enough now :-) 

GTG?

will:
Thanks!

bcardiff:
Awesome!

Whenever you are ready I think that publishing a 0.13.0 we will be ready for a little announcement in crystal's blog. 

And related to the newest api, you can link the following resources at the readme if you want:

* https://crystal-lang.org/docs/database/
* https://crystal-lang.org/docs/database/connection_pool.html
I hope to write one regarding transactions soon.

Thanks again and sorry to bother you these last weeks.

will:
@bcardiff released 0.13.0 thanks. No bother at all

PR #72, opened 2016-12-17
Update to crystal-db ~> 0.3.1

zlwaterfield:
crystal-db 0.3.1 was released 

Update to keep in line with other drivers for https://github.com/topaz-crystal/topaz


zlwaterfield:
I see you already have the update in https://github.com/will/crystal-pg/pull/71 so ill close this.

PR #76, opened 2016-12-26
Update CHANGELOG

bcardiff:


will:
hah whoops

bcardiff:
Not sure about your timezone though :-p 
To make "contribs" from mobile is a bit too much sometimes.

PR #78, opened 2017-01-12
ensure listen/notify spec works in same connection

bcardiff:
While working on crystal-lang/crystal-db#35 I notice that the following spec will break.
This PR fixes things beforehand and it also make sense today.

~~I will send a related PR to discuss listen/notify across other connections.~~ see #79
The former spec worked only because the same connection was been used for listen & notify.


will:
Sorry got this confused with the other one and only saw it today

Issue #56, opened 2016-08-15
Support for LISTEN/NOTIFY


Svenskunganka:
Would be really cool to have support for LISTEN/NOTIFY commands, there's some information about it here:

https://www.postgresql.org/docs/current/static/sql-listen.html
https://www.postgresql.org/docs/current/static/sql-notify.html

Real-time sites could be built in a similar fashion to [this golang example](http://coussej.github.io/2015/09/15/Listening-to-generic-JSON-notifications-from-PostgreSQL-in-Go/) with WebSockets.

will:
You can set the on_notify callback on the connection to process listen notify events. While it is async, there has to be some network traffic going for the events to make it in, such as just an empty select

``` crystal
require "./src/pg"

DB = PG.connect("postgres:///")

DB.on_notification do |note|
  p note
end

DB.exec "listen a"
DB.exec "notify a, 'hello'"

loop do
  sleep 1
  DB.exec "select"
end

# output
# PQ::Notification(@pid=54318, @channel="a", @payload="hello")
```

Svenskunganka:
I missed that completely, thank you for clarifying!

I have a couple of follow-up questions; You said that network traffic has to be active for the events to come in. Is there any particular reason to this? In your example, are notifications received at most every second or are they "instant", even though the empty select statement is only executed once per second?

Thanks!

will:
Right now the driver does not read from the socket unless it is expecting a response from the server, and if it happens to be an unexpected notice, error, or parameter message it process them before going back to what it expects. 

It might be possible to make it instant if there is a way to know when the socket has data, or near instant by spawning something that polls the socket rather than making a complete query, but that would need to make sure not to read a byte in the middle of a query/response flow and get everything out of sync.

This driver is the only serious thing I've done with networking, so I'm not yet sure of the correct approach.

jhass:
I guess the protocol specifies some kind of message/packet format, so that you could continuously read those from a socket? Then you could try to spawn a fiber that continuously reads and distributes to various various channels:

```cr
def poll
  spawn do
    loop do
      data = @socket.read(buffer) # read length header and right length instead
      message = Message.new(data)
      if message.kind.notification?
        @notifications.send message
      else
        @query_results.send message
      end
    end
  end
end

def query
  @socket.write data
  message = @query_results.receive
  # Perhaps data has some tag we can check here to see whether the message is for us
  # and if not return it to the channel
  @query_results.send message unless message.tag == tag
  
  case message.kind
[...]

PR #79, opened 2017-01-12
allow listen/notify to work from different connections

bcardiff:
Right now, there is no clear way to use listen/notify from different connections.

Given a `PG::Connection` there is no way to reach the (protected) `#listen()` method that will spawn a call to `#read_async_frame_loop`.

This PR turns `#listen(*channels : String)` public. It is the minimal change to allow the new spec to pass.

I would go with an implementation that use/add `#listen(*channels : String)` `#unlisten(*channels : String)` to track whether the connection should have an spawned fiber to process the notifications. The `#unlisten` should stop de spawned fiber probably when it is no longer needed.

I am not sure how that impact with notifications received via `Frame::NotificationResponse`

So, rather a full PR this open a discussion + offers a minimal implementation to deal with  listen/notify from different connections.

### DIFF
--- a/spec/pg/connection_spec.cr
+++ b/spec/pg/connection_spec.cr
@@ -46,6 +46,33 @@ describe PG::Connection, "#on_notification" do
     last_note.not_nil!.channel.should eq("somechannel")
     last_note.not_nil!.payload.should eq("do a thing")
   end
+
+  it "does listen/notify from other connection" do
+    last_note = nil
+
+    listening = Channel(Nil).new
+    notified = Channel(Nil).new
+    stop = Channel(Nil).new
+
+    with_db do |db|
+      spawn do
+        db.using_connection do |conn|
+          conn.on_notification { |note| last_note = note; notified.send(nil) }
+          conn.listen("somechannel")
+          listening.send(nil)
+          stop.receive
+        end
+      end
+
+      listening.receive
+      db.exec("notify somechannel, 'do a thing'")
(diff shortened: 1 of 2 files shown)


will:
When the connection is in the special listening mode, it can't do regular queries. I made it so you have to set up that connection from a separate method, rather than turning a connection into that mode, to try and prevent it from ever being used for normal queries. 

Alternatively I suppose the connection could check every query if it was in the listening mode and then throw a runtime error, but I'm not sure if that's better.

will:
It's also possible to get a connection to do both, but it'd be a lot of work: https://github.com/will/crystal-pg/issues/56#issuecomment-240246679

bcardiff:
I see. I don't have a use case right now but from an api I think is better for the runtime exception.

The this is that if the _change connection mode_ is not done, when crystal-lang/crystal-db#35 is fixed the code at [ListenConnection](https://github.com/will/crystal-pg/blob/master/src/pg.cr#L20) will stop working. 

We could add a `Database#get_connection` that can work without a block and will be user's responsibility to release it. Or implement crystal-lang/crystal-db#13. 

Do you prefer any course of action?

bcardiff:
`Database#checkout` `Connection#release` will be added by https://github.com/crystal-lang/crystal-db/pull/38 . Once merged, I will submit another PR to update ListenConnection so it won't break.

So the issue about how to handle listen/notify from connection of the pool can be handled later.

PR #86, opened 2017-03-10
Use vm based ci

waghanza:
+ I've noticed that `crystal-pg` was using `postgresql` **9.4** libraries on all tests suites (even when specified `9.1`), so I change to **sudo-based**
+ I use socket connections instead of IP based, due to an travis error (I think) => reported
https://github.com/travis-ci/travis-ci/issues/7449
+ I add test env using `postgresql `**9.5** / **9.6**

PS : 
Travis feature
~~~yaml
addons
~~~
is deprecated https://docs.travis-ci.com/user/addons/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,21 +2,26 @@ dist: trusty
 
 sudo: required
 
-addons:
-  postgresql: "9.4"
+services:
+  - postgresql
 
 language: crystal
+
 crystal:
   - latest
+
 env:
+  - TRAVIS_POSTGRESQL_VERSION=9.6
+  - TRAVIS_POSTGRESQL_VERSION=9.5
   - TRAVIS_POSTGRESQL_VERSION=9.4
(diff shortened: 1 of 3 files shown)


will:
Thanks. I will merge this, but I'm somewhat worried about the error that the non-sudo was exposing. I don’t understand it at all, so I'm unsure if it's travis's fault or if it's some bug in crystal-pg or crystal.

waghanza:
I'm not aware of all `travis` stuff but I have made an issue on the tracker.

I can test it on `circleci` if you want.

waghanza:
@will Seems to work on circle without any modification https://circleci.com/gh/waghanza/crystal-pg/8 but cirlcleci does not allow multi-version testing (as far as I know)

waghanza:
https://github.com/travis-ci/travis-ci/issues/7449

will:
Thanks!

PR #87, opened 2017-03-20
Update to crystal-db 0.4.0

bcardiff:
* Use `DB.connect` in ListenConnection to avoid creating a connection pool.
* Updates to use db 0.4.0.

### DIFF
--- a/shard.yml
+++ b/shard.yml
@@ -4,4 +4,4 @@ version: 0.13.2
 dependencies:
   db:
     github: crystal-lang/crystal-db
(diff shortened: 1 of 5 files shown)


bcardiff:
@will this PR does not force a bump in the minor version. It can be just a bump in the patch if you prefer.

With DB.connect, the user is able to use the connection pool he please, but without the built in retry logic crystal-db offers.

will:
Thanks, merged and released 0.13.3

Issue #88, opened 2017-03-25
Decoding types with unstable OIDs


bjeanes:
Hi @will (hope you're well!)

I'm working on a Crystal side-project at the moment that uses `crystal-pg`. It's working with a legacy database which has an eclectic mix of types that would have different OIDs from DB to DB and thus can't have statically defined decoders, as `crystal-pg` does.

I'm thinking of course of `citext`, `hstore`, and, in my case anyway, custom `DOMAIN`s based on other types.

I've been looking through the source code for `crystal-pg` (very readable, btw) and it occurs to me that to support something like `hstore`, the decoder map would have to be built up on a per-connection basis (vs globally) by querying `pg_type` et al as [Ruby's PG gem does](https://bitbucket.org/ged/ruby-pg/src/8a15026cfe7792eb1aff5cb77fefe109e8dbf6aa/lib/pg/basic_type_mapping.rb?at=default&fileviewer=file-view-default#basic_type_mapping.rb-106:107).

I've not played with the `pg_type` table much, but I wonder if the [`typcategory` column](https://www.postgresql.org/docs/8.4/static/catalog-pg-type.html#CATALOG-TYPCATEGORY-TABLE) could be used as a hint for a fallback decoder to use where a specialised one wasn't available. That might allow `citext` (or for e.g. an `DOMAIN` "`email`" based on `citext`) to naturally be decoded as would `text`.

I whipped up a query to get familiar with what the types in my DB looked like and the relationships between them, and I think something like this could work.

Query:

``` sql
WITH

categories(typcategory, category) AS (
  VALUES
    ('A', 'Array')
  , ('B', 'Boolean')
  , ('C', 'Composite')
  , ('D', 'Date/time')
  , ('E', 'Enum')
  , ('G', 'Geometric')
  , ('I', 'Network address')
  , ('N', 'Numeric')
  , ('P', 'Pseudo-types')
  , ('S', 'String')
  , ('T', 'Timespan')
  , ('U', 'User-defined')
  , ('V', 'Bit-string')
  , ('X', 'unknown')
),

types AS (
  SELECT
                                        oid
  , typtype                             "type"
  , typnamespace                        namespace_oid
  , typnamespace::regnamespace          namespace
  , typname                             "name"
  , typcategory                         category
  , (typcategory || ' - ' || category)  category_name
  , typbasetype                         base_oid
  , typelem                             element_oid
  , typarray                            array_oid
  , typdelim                            array_delimeter
[...]

PR #89, opened 2017-03-25
Per-connection decoders with PoC `citext` and `hstore` support

bjeanes:
This is a follow up to my thoughts in #88.

## Summary

Reading from a result set picks a decoder via a connection-local decoder map instead of the global list. Misses on the local map fallback to the existing global map of handlers.

When a connection is first initialised, it is immediately used to interrogate `pg_type` and expand on the statically-defined global decoder map.

This mechanism is used to implement `citext` support.

<details>
<summary>Notes</summary>

* This will cause types to be determined multiple types in a connection pooling scenario. It may instead be preferable to do this once per `DB`. However, given that connections may or may not be pooled, the current approach seemed the most prudent, at least from a proof-of-concept standpoint.
* Changes to the types during a connection won't be visible (e.g. `CREATE EXTENSION`). There's a few options here, but they don't seem deeply important to address now. It does mean the test for `citext` has to jump through some hoops, though.
* The statically-defined global map of decoder handlers is depended upon to interpret the result set from querying `pg_type`.
* I've left more comments and commented-out code than I usually would as the point of this PR is foremost to communicate the idea. E.g. A small snippet of code shows how a specialised `HstoreDecoder` could be wired up, too.

</details>

### DIFF
--- a/src/pg/decoder.cr
+++ b/src/pg/decoder.cr
@@ -302,12 +302,7 @@ module PG
     end
 
     TYPE_SQL = %q(
-      SELECT
-        /* Avoid search_path woes by qualifying before cast... */
-        (typnamespace::regnamespace::text || '.' || typname)::regtype::oid
-                                              oid
-      , typname                               "name"
-      , typcategory                           category
+      SELECT oid, typname, typcategory
       FROM pg_type
       WHERE typisdefined = 't'
         AND typtype IN ('b', 'd'))
--- a/src/pg/decoder.cr
+++ b/src/pg/decoder.cr
@@ -139,7 +139,7 @@ module PG
             length = read_u32(io)
             key = string_decoder.decode(io, length)
             length = read_u32(io)
-            if length == UInt32::MAX
+            if length == -1
               hash[key] = nil
             else
               value = string_decoder.decode(io, length)
--- a/src/pg/decoder.cr
+++ b/src/pg/decoder.cr
@@ -300,6 +300,8 @@ module PG
       @@decoders[oid]
     end
 
+    # Globally registers a `Decoder` instance to handle type specified by
+    # provided OID.
     def self.register_decoder(decoder, oid)
       @@decoders[oid] = decoder
     end
--- a/src/pg/decoder.cr
+++ b/src/pg/decoder.cr
@@ -327,9 +327,7 @@ module PG
 
     # Builds a `DecoderMap` of all types visible on `connection` that are not
     # already statically known.
-    def self.build_decoder_list(connection : PG::Connection) : DecoderMap
-      decoders = DecoderMap.new { |_, oid| from_oid(oid) }
-
+    def self.register_connection_decoders(connection : PG::Connection) : Void
       types = connection.query_all(TYPE_SQL, as: {UInt32, String, Char})
       types.each do |oid, name, category|
         oid = oid.to_i32 # Query execution if I read straight to Int32 :\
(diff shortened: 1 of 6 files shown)


bjeanes:
Added an `HstoreDecoder` and param handling but it's probably pretty naïve.

bjeanes:
Further thoughts:

* A lot of types will use the same decoder and as more types are supported, the number of these decoder instances will grow with the connections. They aren't stateful so perhaps stand to be made singletons.
* This first pass just demonstrates loading a fallback decoder based on the type category and loading one based on the type name. However, for other `DOMAIN`s actually looking up the decoder based on the base type's OID would be good.
* Array decoding doesn't (yet) benefit from this approach. `pg_type` does contain enough information that we could add extra array decoders. Though I don't yet know enough about Crystal to see how this dynamic approach could support the recursive type definition employed now.
* Is there any reason `read_array` and `read` approaches can't be unified? Array entries in `pg_type` include OID for the elements' type, which could be used to pass in the element decoder.
* This code still needs a bit of factoring (I don't think this new code all belongs in `Decoders` directly—probably in a child namespace).

will:
I need to think more on your questions. I'm at a postgres conf this week (and giving a talk including the protocol parts of this project 😀), so I'll dive in next week. Thanks for putting this together!

bjeanes:
No worries! I actually have another approach that may generalise better and
handle more cases so may have another similar PR next week anyway (if i can
spare the time–otherwise I'll leave a describing comment instead)

will:
Hey @bjeanes did you have any progress on the other approach?

bjeanes:
@will ah no I got sidetracked with other things.

I was basically wanting to turn the `DecoderMap` as `Hash(OID, Decoder)` to `TypeMap` as `Hash(OID, PG::Type)`. `PG::Type` is a struct which includes a (singleton) reference to decoder but also also more information from the `pg_type` table. I don't know if this is actually valuable but I have a hunch that it could make array decoding less of a special-case and allow for some sane assumptions for newly encountered types based on the category (geometric, numeric, etc).

If not that approach, I think my actual PoC here could be factored to be less ad-hoc as I kinda just dumped some code in the existing decoder code. 

I dunno... what do you think? Did you have any imagined direction when you've contemplated per-connection type support? Anything you'd want to see to have a this or a derivative of this be merged?

bjeanes:
Oh I also think there's more opportunity for composite or higher-order decoders to leverage the primitive decoders and perhaps for some symmetry between encoding/decoding, which might make it a bit clearer when adding support for future types.

src/pg/decoder.cr
```diff
@@ -294,12 +317,56 @@ module PG
...
+      BIT_STRING   = 'V'
+      UNKNOWN      = 'X'
+    end
+
+    # Builds a `DecoderMap` of all types visible on `connection` that are not
+    # already statically known.
+    def self.register_connection_decoders(connection : PG::Connection) : Void
+      types = connection.query_all(TYPE_SQL, as: {UInt32, String, Char})
+      types.each do |oid, name, category|
+        oid = oid.to_i32 # Query execution if I read straight to Int32 :\
```
bjeanes: Oops this comment should read "query execution hangs if...".

That was a reminder to me to ask @will why that is? I was fighting a lot of mysterious hangs if I didn't get things right. It can be hard to debug sometimes. Any advice?

bjeanes:
I've been thinking about picking back up my side project that needed this functionality and drove this PR. @will are you still working on this lib and if so what do you see as the path forward for hstore/citext support?

will:
Yeah, this is still a project. I think the only way for any sort of extension data type where the oid changes from database to database will require a per-connection mapping. If you're interested in working on this more, I'm happy to help if you run into anything.

PR #96, opened 2017-04-15
Use db shared specs

bcardiff:
crystal-db 0.4.1 includes some common specs across drivers. This is a minimum usage of those.

It could include other datatypes that crystal-pg supports. It apart from testing literals in a similar fashion that in decoder_spec.cr, this shared specs make some heavy usage of many overloads to ensure all db api works with all sample value data types.

Unfortunately for the specs to pass we need `$ crystal spec --no-debug` until the crystal version is released. That is the sole reason why the `[WIP]` on my side.

FYI Inside the `DB::DriverSpecs(DB::Any).run` you can: 

```crystal
DB::DriverSpecs(DB::Any).run
   it "..." do |db|
     # use a spec with an empty db :-)
   end
end

### DIFF
--- a/spec/db_spec.cr
+++ b/spec/db_spec.cr
@@ -1,0 +1,75 @@
+require "./spec_helper"
+require "db/spec"
+
+private def cast(expr, sql_type)
+  if sql_type
+    "#{expr}::#{sql_type}"
+  else
+    expr
+  end
+end
+
+private def bytes(*values)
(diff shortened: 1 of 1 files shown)


bcardiff:
Rebased on master. @will what do you think about this change? A benefit is that future PRs in crystal-db that requires specific behaviour on drivers like https://github.com/crystal-lang/crystal-db/pull/156 will get specs for free :-)

will:
Thanks for bringing it back up!

Issue #102, opened 2017-07-13
Nullable Text Field Record Causing Exception


woodzer:
If you create a simple table containing a column with a type of ```text``` that accepts null values then you get an exception and stack trace from the following code if there are any records in the table...

```crystal
require "db"
require "pg"
require "./db_issue/*"

module DbIssue
  DB.open("postgres://user_name@localhost/database_name") do |connection|
    connection.query("select * from test_table") do |resultset|
    end
  end
end
```

Obviously you'll need to replace ```user_name``` and ```database_name``` with the appropriate values in this code. The stack trace generated (on OSX) is...

```
Index out of bounds (IndexError)
0x103ca4515: *CallStack::unwind:Array(Pointer(Void)) at ??
0x103ca44b1: *CallStack#initialize:Array(Pointer(Void)) at ??
0x103ca4488: *CallStack::new:CallStack at ??
0x103c936d5: *raise<IndexError>:NoReturn at ??
0x103cc776a: *Slice(UInt8)@Slice(T)#+<Int32>:Slice(UInt8) at ??
0x103d5214d: *Socket+@IO::Buffered#skip<Int32>:Nil at ??
0x103d4c932: *PQ::Connection#skip_bytes<Int32>:Nil at ??
0x103d6c875: *PG::ResultSet#skip:Int32 at ??
0x103d6c2fd: *PG::ResultSet#do_close:Nil at ??
0x103d6c22d: *PG::ResultSet@DB::Disposable#close:(Bool | Nil) at ??
0x103c88c19: __crystal_main at ??
0x103c9d1f8: main at ??
```

This doesn't get generated if the nullable column actually has a value or if the table has no records in it.

will:
I was able to figure out that this also will happen with a `DB.mapping` if there are two, but not one, nulls. If there is just one, the crystal-db wrapper appears to notice it.

Also after this happens, it leaves the connection in a bad state and normal queries no longer work

```crystal 
require "pg"
PGDB = DB.open("postgres:///")

struct Foo
  DB.mapping(a: Int32, b: Int32)
end

def capture(title)
  puts "\n\n#{title}"
  begin
    yield
  rescue e
    puts e.message
    p e.backtrace
  end
end

def normal
  capture("normal query") do
    p PGDB.query_all("select 1 as id union select null as id", &.read(Int32?))
  end
end

normal

capture("mapping with no nulls") do
  p Foo.from_rs(PGDB.query("select 1 as a, 1 as b"))
end

normal

[...]

PR #107, opened 2017-09-01
Fix skipping null values

ZeWebDev:
crystal-pg would hang indefinitely or throw an error when there were unread null values.
The skip function didn't skip bytes when the column byte size was 1 but it should be -1 which is when it's null.

Fixes
https://github.com/will/crystal-pg/issues/102
https://github.com/crystal-lang/crystal-db/issues/59


will:
Thanks!

Issue #112, opened 2017-09-21
Returning last inserted id


vladfaust:
Refer to https://github.com/will/crystal-pg/blob/800540888edf4012d2fdaba5d508665b462700dc/src/pg/statement.cr#L39

But there is an option with `RETURNING`; maybe this could be implemented?

will:
I'm not sure how the driver here would know to add `RETURNING` automatically, because not every statement is an insert, and not every id column is necessarily called `"id"`

Also, it looks like it's used in this struct https://github.com/crystal-lang/crystal-db/blob/385cf70a8a5d41374cce988168e82495743ee459/src/db.cr#L80 But I often have tables where the ID is not an Int64, often they're UUIDs or something else, so I'm not sure it's generalizable at all. This might be something that comes from mysql or something?

vladfaust:
@will I agree that `last_insert_id : Int64` is idiomatically wrong. Maybe it worths opening an issue in [crystal-db](https://github.com/crystal-lang/crystal-db)? Like, change to `Int64 | String`?

I'm not sure, but aren't you able to check whether is `RETURNING` present in a query? If yes, maybe parse the first value... It add some covenants, though.

vladfaust:
`rows_affected` equals to 0 too.

nicferrier:
I can't even get RETURNING to work, the code gets tangled up. It looks like support was there in the frame.cr code once (DataRow) but it doesn't work now.

vladfaust:
So, could anyone make `RETURNING` work?

RX14:
`RETURNING` has always worked for me

vladfaust:
Hm, yes. It works if doing `db.query("INSERT ... RETURNING *")`... So the issue regrads to `ExecResult`.

RX14:
What?

RX14:
Thats just how postgres works. `ExecResult` is useless for postgres. What if your primary key wasn't an `Int64`?

vladfaust:
Yep. Closing this one then.

waghanza:
Does `RETURNING` works for `exec` ? or only for `query` ?

PR #121, opened 2017-12-30
#112 attempted fix

nicferrier:
I was having problems with RETURNING statements (like an INSERT returning the id of the created row) but it turned out that it was just a problem with async notifications breaking the flow as illustrated.

Basically, what I was getting was notices from a trigger function that happen on the insert. These appear before the DataRow that encapsulate the RETURNING.

Maybe this is a bit messy. But it works.

I suspect that there's a much better way of dealing with the async messages.

### DIFF
--- a/src/pq/connection.cr
+++ b/src/pq/connection.cr
@@ -267,6 +267,15 @@ module PQ
 
     def read_next_row_start
       type = soc.read_char
+
+      if type == 'N'
+        frame = read_one_frame type
+        async_frames = handle_async_frames frame
+        if async_frames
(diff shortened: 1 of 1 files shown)


will:
Thanks for sending this in. I'm traveling this week and won't be able to properly take a look at this before the new year.

Could you please also add a test that fails without this change to pq::connection and passes with it to make sure it doesn't regress again in the future?

waghanza:
@nicferrier Time to complete this `PR` ?

jgaskins:
Looks like this has been implemented in #163.

PR #123, opened 2017-12-31
Update to use crystal-db ~> 0.5.0

### LINKED ISSUE #122: shards update problem when using both crystal-pg and crystal-mysql
hello

I was working on https://github.com/imdrasil/jennifer.cr/pull/84 and just realised that **crystal-mysql** and **crystal-pg** can't be installed side-by-side any more (like how ORMs usually do it, at least in development mode). that is because **crystal-mysql** requires **crystal-db** `~> 0.5.0`, while **crystal-pg** is looking for `~> 0.4.1` so there's no common ground. the new **crystal-db** and **crystal-mysql** versions have been released together yesterday so this is quite fresh.

could you please update the **crystal-db** expectation to 0.5.0 (or at least allow that one as well, like `~> 0.4`, though I'm not sure if the project is strict about semantic versioning)?

bcardiff:
Fixes #122 .
@will a new crystal-db was released for 0.24, no changes in the api.

I realized later that that bump was not really need. I'll try to be more careful next time.

### DIFF
--- a/shard.yml
+++ b/shard.yml
@@ -4,4 +4,4 @@ version: 0.14.0
 dependencies:
   db:
     github: crystal-lang/crystal-db
-    version: ~> 0.4.1
+    version: ~> 0.5.0


will:
Thanks. What version number do you think this should go to with this?

bcardiff:
I prefer to bump a minor, so `0.15.0`, to leave room for future patches in previous version, although probably wont apply. I just think it's a good practice.

But given that `0.14.0` is already targeted to crystal 0.24, it can be `0.14.1`.

will:
Makes sense. I really want to figure out (or accept a patch that adds) the per-connection type maps for extension oids like hstore, citext, etc, then make this 1.0. Then all the version numbers will make more sense with semver.

RX14:
@will thing is, you can't make `1.0` guarantees while crystal continues to change under your feet. I personally don't think that a `1.0` library can exist before crystal is `1.0`.

will:
That's a good point

PR #130, opened 2018-01-21
decode uuid with builtin UUID type

pynixwang:
more clear, but use more memory.

### DIFF
--- a/src/pg/decoder.cr
+++ b/src/pg/decoder.cr
@@ -1,4 +1,5 @@
 require "json"
+require "uuid"
 
 module PG
   alias PGValue = String | Nil | Bool | Int32 | Float32 | Float64 | Time | JSON::Type | PG::Numeric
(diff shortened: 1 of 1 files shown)


pynixwang:
only one io.

RX14:
It doesn't use more memory - `UUID` is a struct.

pynixwang:
@RX14 but allocate UInt8[16]

RX14:
That's also on the stack. Llvm will likely inline the constructor and to_s and elide the memory copy and you get the same code.

pynixwang:
@RX14 so.

will:
Hi, I appreciate the refactor, and agree it is easier to reason about. It does however appear to be measurably slower, unless I'm doing the benchmark wrong. 


```
/tmp ➤ crystal build --release a.cr
/tmp ➤ ./a
old   5.76M (173.65ns) (± 4.32%)       fastest
new   4.24M (235.96ns) (± 4.67%)  1.36× slower
/tmp ➤ ./a
old   5.74M (174.08ns) (± 4.90%)       fastest
new    4.2M (237.98ns) (± 5.86%)  1.37× slower
/tmp ➤ ./a
old   5.87M (170.44ns) (± 3.88%)       fastest
new   3.96M ( 252.7ns) (±10.76%)  1.48× slower
```

``` crystal
require "benchmark"
require "uuid"

def old(io)
  bytes = uninitialized UInt8[6]

  String.new(36) do |buffer|
    buffer[8] = buffer[13] = buffer[18] = buffer[23] = 45_u8

    slice = bytes.to_slice[0, 4]

    io.read(slice)
    slice.hexstring(buffer + 0)

    slice = bytes.to_slice[0, 2]

    io.read(slice)
    slice.hexstring(buffer + 9)

    io.read(slice)
    slice.hexstring(buffer + 14)

    io.read(slice)
    slice.hexstring(buffer + 19)

    slice = bytes.to_slice
    io.read(slice)
    slice.hexstring(buffer + 24)

    {36, 36}
  end
end

def new(io)
  bytes = uninitialized UInt8[16]
  io.read(bytes.to_slice)
  UUID.new(bytes).to_s
end

STR = "}a\xD5H\u0012LK8\xBC\u0005ϻ\x88\xCF\xD1\xD1"

def old_go
  old(IO::Memory.new(STR))
end

def new_go
  new(IO::Memory.new(STR))
end

raise "broken" unless old_go == new_go == "7d61d548-124c-4b38-bc05-cfbb88cfd1d1"

Benchmark.ips do |x|
  x.report("old") { old_go }
  x.report("new") { new_go }
end
```

Since this is covered by tests, and doesn't change often, I'd prefer to keep it faster.

RX14:
However, returning UUIDs as the new `UUID` struct would be something to discuss... Not sure it's worth it.

pynixwang:
@RX14 UUID always use as `String`

RX14:
?

pynixwang:
some property such as `version` may never be used

PR #138, opened 2018-03-17
Support connection negotiation without BackendKeyData frame

RX14:


src/pq/connection.cr
```diff
@@ -235,11 +232,16 @@ module PQ
       auth_frame = expect_frame Frame::Authentication
       handle_auth auth_frame
 
-      key_data = expect_frame Frame::BackendKeyData
-      @pid = key_data.pid
-      @secret = key_data.secret
```
will: If we don’t store the pid and secret, it's impossible to cancel queries

RX14:
Ping

will:
Thanks, sorry for having this drop off my radar again

PR #141, opened 2018-05-13
fix compilation in crystal master

kostya:


bcardiff:
Hi @kostya the #131 is already taking care of this and some other upgrades to crystal 0.25.
If you are using crystal master you might find it useful.

Thanks!

PR #151, opened 2018-08-26
Check connection client encoding is UTF8, case-insensitive

yumoose:
Fixes connections with encoding set as lowercase `utf8` as opposed to `UTF8` that still have compatible encoding.

Notably, this prevents an error being raised when using [PgBouncer](https://pgbouncer.github.io/) in Heroku for [client-side connection pooling](https://devcenter.heroku.com/articles/on-dyno-postgres-connection-pooling).

### DIFF
--- a/src/pq/connection.cr
+++ b/src/pq/connection.cr
@@ -207,7 +207,7 @@ module PQ
       @server_parameters[frame.key] = frame.value
       case frame.key
       when "client_encoding"
-        if frame.value != "UTF8"
+        if frame.value.upcase != "UTF8"
           raise ConnectionError.new(
             "Only UTF8 is supported for client_encoding, got: #{frame.value.inspect}")
         end


will:
Great! Thank you.

Indribell:
We just ran into this same issue
```
Caused by: Only UTF8 is supported for client_encoding, got: "utf8" (PQ::ConnectionError)
```

Please release a updated version because currently version 0.15 from Jun 15 ( predating this change ) is the latest officially that is available for downloaded with shards.

straight-shoota:
@Indribell You can tell shards to use master (or a specific version with this patch) by adding `branch: master` (or `commit: <commit-hash>`) to your `shard.yml`.

Indribell:
Thanks you @straight-shoota. This information is not very well represented in the Crystal manual. Most that i read refers to versions unfortunately.

PR #152, opened 2018-09-27
Enable clientcert authentication

sanderhahn:
Allow clientcert authentication using query parameters that resemble the parameter key words (`sslcert`, `sslkey`, `sslrootcert`): https://www.postgresql.org/docs/current/static/libpq-connect.html#LIBPQ-PARAMKEYWORDS.

### DIFF
--- a/src/pq/connection.cr
+++ b/src/pq/connection.cr
@@ -54,6 +54,15 @@ module PQ
       if serv_ssl
         ctx = OpenSSL::SSL::Context::Client.new
         ctx.verify_mode = OpenSSL::SSL::VerifyMode::NONE # currently emulating sslmode 'require' not verify_ca or verify_full
+        if sslcert = @conninfo.sslcert
+          ctx.certificate_chain = sslcert
+        end
+        if sslkey = @conninfo.sslkey
+          ctx.private_key = sslkey
+        end
+        if sslrootcert = @conninfo.sslrootcert
+          ctx.ca_certificates = sslrootcert
+        end
         @soc = OpenSSL::SSL::Socket::Client.new(@soc, context: ctx, sync_close: true)
       end
 
--- a/src/pq/conninfo.cr
+++ b/src/pq/conninfo.cr
@@ -19,11 +19,17 @@ module PQ
     # The password. Optional.
     getter password : String?
 
-    # SSLmode, either :prefer or :require
-    #
-    # If set to `:prefer`, ssl is attempted. If it is `:require` and ssl is not
-    # established, an error is raised
+    # The sslmode. Optional (:prefer is default).
     getter sslmode : Symbol
+
+    # The sslcert. Optional.
+    getter sslcert : String?
+
+    # The sslkey. Optional.
+    getter sslkey : String?
+
+    # The sslrootcert. Optional.
+    getter sslrootcert : String?
 
     # Create a new ConnInfo from all parts
     def initialize(host : String? = nil, database : String? = nil, user : String? = nil, @password : String? = nil, port : Int | String? = 5432, sslmode : String | Symbol? = nil)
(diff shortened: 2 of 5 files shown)


src/pq/connection.cr
```diff
@@ -54,6 +54,18 @@ module PQ
...
+          ctx.certificate_chain = sslcert
+        end
+        sslkey = @conninfo.sslkey
+        if sslkey
+          ctx.private_key = sslkey
+        end
+        sslrootcert = @conninfo.sslrootcert
+        if sslrootcert
+          ctx.ca_certificates = sslrootcert
+        end
```
will: This can be a little more concise if they read something like 
```crystal 
ctx.ca_certificates = @conninfo.sslrootcert if @conninfo.sslrootcert
```

I haven't done much with cert auth myself, so I don’t know if all 3 are always required. If so, it could instead be one if block that sets all 3 rather than 3 separate conditionals.

src/pq/conninfo.cr
```diff
@@ -33,6 +42,9 @@ module PQ
       @user = default_user user
       @port = (port || 5432).to_i
       @sslmode = default_sslmode sslmode
+      @sslcert = ENV["PGSSLCERT"]?
+      @sslkey = ENV["PGSSLKEY"]?
+      @sslrootcert = ENV["PGSSLROOTCERT"]?
```
will: I think this would become the only place where the library is directly looking at environment variables. Would it be possible to make this be an argument, and have people pick how they get this info into the library, be it env vars, or some other means. I realize that these are probably the env vars that libpq uses, but it would be nice to be a little more flexible.

It may also make sense to make a struct that holds these three values, if that makes passing it around easier.

src/pq/connection.cr
sanderhahn: New to Crystal so still looking of the right way to formulate code in an idiomatic way.

src/pq/conninfo.cr
sanderhahn: Thought that using the predefined PostgreSQL environment variables could a be sensible default because it is compatible with `libpq` https://www.postgresql.org/docs/current/static/libpq-envars.html. However i understand that this should probably also include the other `PG*` environment variables. At the moment i build an url like this:

```crystal
class DBUri < URI
  def initialize(**args)
    self.scheme = "postgres"
    self.user = ENV["PGUSER"]
    self.password = ENV["PGPASSWORD"]?
    self.host = ENV["PGHOST"]?
    self.port = ENV["PGPORT"]?.try &.to_i
    self.path = "/" + ENV["PGDATABASE"]
  end
end
```

db = DB.open DBUri.new
``` 

However i am unsure how to pass in the certificate information because all the `crystal-db` api's seem to use urls, such as the `DB.open` interface http://crystal-lang.github.io/crystal-db/api/0.5.0/DB.html#open%28uri%3AURI%7CString%29-class-method.

Comment (sanderhahn) [reply src/pq/conninfo.cr]: Used this article for my setup https://www.gab.lc/articles/postgresql_with_ssl/, but i am unsure how i should add these steps to the spec...

File: src/pq/connection.cr (hunk 1 above)
Comment (sanderhahn) [reply src/pq/connection.cr]: Changed the code a bit to remove the extra assignment lines.

**Comment by sanderhahn** [comment]
Okay tried to remove the environment variables and use the query parameters. In this version its possible to initialize using:

```crystal
class DBUri < URI
  def initialize(**args)
    self.scheme = "postgres"
    self.user = ENV["PGUSER"]
    self.password = ENV["PGPASSWORD"]?
    self.host = ENV["PGHOST"]?
    self.port = ENV["PGPORT"]?.try &.to_i
    self.path = "/" + ENV["PGDATABASE"]
    params = HTTP::Params.build do |form|
      form.add "sslmode", "verify-full"
      form.add "sslcert", ENV["PGSSLCERT"]
      form.add "sslkey", ENV["PGSSLKEY"]
      form.add "sslrootcert", ENV["PGSSLROOTCERT"]
    end
    self.query = params.to_s
  end
end

db = DB.open DBUri.new
```

sanderhahn:
Added integration setup/test for clientcert which seems to pass. However somehow the test now fails to authenticate the wrong `md5` password. Giving up for now, maybe you can see what is going wrong?

src/pq/conninfo.cr
sanderhahn: Now using query parameters to pass in the ssl parameters instead of relying environment variables.

sanderhahn:
:)

PR #153, opened 2018-10-07
Make column_type_oid return oid and introduce PG::Type::*OID constants

sanderhahn:
Was looking for a way to reflect on the type of the column returned by a `ResultSet` and found the `column_type` method. However using it resulted in a compile error, maybe because of changes in how decoders are registered. This PR will make it return `oid` information from the `pg_type` table.


sanderhahn:
Some oids can be user defined such as domains and enums. So i thought maybe return the oid itself, because domains and enums are not hardcoded.

sanderhahn:
Changed the `column_type` to return an `oid` value and introduced constants for the `pg_types` for easy reference.

PR #160, opened 2019-03-28
Fix encoding string arrays with special characters

### LINKED ISSUE #159: Array encode does not escape control characters
`PQ::Param.encode_array` needs to properly escape items:

```cr
DB.open(ENV["DATABASE_URL"]) do |db|
  db.query_one "SELECT $1::text[]", [["baz, bar"]], as: Array(String) # => ["baz", "bar"]
  db.query_one "SELECT $1::text[]", [["foo}"]], as: Array(String) # malformed array literal: "{foo}}"
end

straight-shoota:
Fixes #159

### DIFF
--- a/spec/pg/encoder_spec.cr
+++ b/spec/pg/encoder_spec.cr
@@ -18,30 +18,46 @@ end
 
 describe PG::Driver, "encoder" do
   test_insert_and_read "int4", 123
+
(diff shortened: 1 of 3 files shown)


will:
Thanks for fixing this and especially the tests!

PR #161, opened 2019-03-28
Fix version parser and note PostgreSQL 11 compatibility

straight-shoota:


README.md
```diff
@@ -55,7 +55,7 @@ PG_DB.query_one("select '{hello, world}'::text[]", &.read(Array(String))
 ## Requirements
 
 Crystal-pg is [tested on](https://travis-ci.org/will/crystal-pg) Postgres
-versions 9.2 through 9.6. Since it uses protocol version 3, older versions probably also
+versions 9.2 through 11. Since it uses protocol version 3, older versions probably also
```
will: On traivs, it's only doing 9.2 to 9.6, so this was saying that is what is tested. The driver supports through 11 and the foreseeable future until they change the protocol, but this line was about what was tested. So would you want to instead change this to add a sentence like that with the distinction between testing and support?

Travis may have other versions now, but I know for a while you'd have to manually downlaod other versions yourself and they'd lag pg support by years. (But also the new owners of travis fired nearly everyone so maybe it's time to switch CI…)

straight-shoota: Yeah, I've missed that. I primarily wanted this to express that it works with most recent PostgreSQL as well, not only with a two-year-old release. I'll change the phrasing as suggested.

Having actual testing for 11 would obviously be great. I suppose a relatively easy implementation would be a multi stage workflow building the spec and running the binary in several docker images providing different PostgreSQL Versions.

will:
Thanks!

PR #162, opened 2019-03-28
Fix reading null when expecting array

### LINKED ISSUE #157: Trying to read a empty array causes infinite hang/sleep
When trying to read/map a array but receiving a empty cell (`null`), it'll just hang endless instead of simply raising a exception.
```cr
require "pg"

DATA = PG.connect(ENV["POSTGRES"])

DATA.exec("DROP TABLE IF EXISTS test;")
DATA.exec("CREATE TABLE test (text_array text[]);")

struct Test
  DB.mapping(
    text_array: Array(String)
  )
end

DATA.exec(<<-SQL)
    INSERT INTO test(text_array) VALUES ('{"hi", "hey"}');
SQL

def query
  DATA.query_all("SELECT * FROM test", as: Test)
end

pp query

DATA.exec("INSERT INTO test VALUES ('{}');")

pp query

DATA.exec("INSERT INTO test VALUES (null);")

# When trying to query now, it'll simply hang infinite
pp query
```

Making it nil-able, gets rid of the error of course. 
```cr
DB.mapping(
    text_array: Array(String)?
)
```

I wasn't sure if this was a crystal-pg or crystal-db issue, but figured more likely crystal-pg.

straight-shoota:
Fixes #157

### DIFF
--- a/spec/pg/decoders/array_decoder_spec.cr
+++ b/spec/pg/decoders/array_decoder_spec.cr
@@ -60,4 +60,10 @@ describe PG::Decoders do
   test_decode "int8 array", "$${1,2}$$::int8[]", [1, 2]
   test_decode "float4 array", "$${1.1,2.2}$$::float4[]", [1.1_f32, 2.2_f32]
   test_decode "float8 array", "$${1.1,2.2}$$::float8[]", [1.1_f64, 2.2_f64]
+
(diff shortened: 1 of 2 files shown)


will:
Thanks, this was merged but I did it as a rebase locally so I supposed github didn't notice.

PR #163, opened 2019-04-11
Fix support for reading a NoticeResponse at row start

straight-shoota:


spec/pg/connection_spec.cr
```diff
@@ -73,3 +73,28 @@ describe PG, "#listen" do
     end
   end
 end
+
+describe PG, "#read_next_row_start" do
+  it "handles reading a notice" do
+    with_connection do |db|
+      db.exec <<-SQL
```
will: can you add `SET client_min_messages TO notice;` to the start here? I think CI has a different setting by default for some reason, and that's why it's failing.

straight-shoota: Yeah, that was it :+1:

will:
Thanks!

PR #167, opened 2019-05-04
Add `Enumerable` channels overload for `PG.connect_listen`

### LINKED ISSUE #166: `PG.connect_listen` should accept `Enumerable` arguments
It's a good practice to write splat methods as overloads for `Enumerable` methods like this:

```crystal
def foo(channels : Enumerable(String))
  pp channels
end

def foo(*channels : String)
  foo(channels)
end

foo("a", "b")   # => {"a", "b"}
foo(["a", "b"]) # => ["a", "b"]
```

As `channels` are called with `.each` @

https://github.com/will/crystal-pg/blob/cafe761ead8f5189fab9de97dec9d05652a305df/src/pg/connection.cr#L45

Then we can safely have `PG.connect_listen(channels : Enumerable(String))` and `PG.connect_listen(*channels : String)`.

vladfaust:
Closes #166

### DIFF
--- a/spec/pg/connection_spec.cr
+++ b/spec/pg/connection_spec.cr
@@ -54,7 +54,7 @@ describe PG, "#listen" do
   it "opens a special listen only connection" do
     got = false
     ch = Channel(Nil).new
-    conn = PG.connect_listen(DB_URL, "foo") do |n|
+    conn = PG.connect_listen(DB_URL, "foo", "bar") do |n|
       got = true
       ch.send(nil)
     end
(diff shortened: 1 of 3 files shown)


will:
Thanks! Can you also add a spec that uses the new signature so that we don’t accidentally break this in the future?

vladfaust:
It's a bikeshedding, since the overloader is literally a one-line proxy.

vladfaust:
I've updates the specs so it accepts multiple arguments. This should be enough.

bcardiff:
@vladfaust IMO it's not bikeshedding. Code that is not called is not analyzed/compiled by crystal so it can go broken undetected easily. Having a call to that method improves a lot the coverage. Even if the implementation is trivial.

Thanks for taking the stab and adding this enhancement.

vladfaust:
You may be right, Brian. Glad I can somehow improve the tool I use in everyday development :slightly_smiling_face:

will:
Thank you!

PR #170, opened 2019-06-04
Surround call PQ::Connection#close with rescue

bcardiff:
* The `PQ::Connection#close` might raise. `#do_close` should never raise
* A call to super to clean up resources was missing


will:
Thanks for this and #169!

PR #171, opened 2019-07-08
Fix IOError on closed connection

omarroth:
If a closed connection is used in a query it will immediately exit with `IO::Error: Closed stream`. Since currently only `EOFError`s are rescued, a closed connection will be re-used in subsequent queries and will always fail.

Please let me know if this is not the correct fix for this or if there is another reason for using `IO::EOFError` and not the parent `IO::Error`.

### DIFF
--- a/src/pg/statement.cr
+++ b/src/pg/statement.cr
@@ -27,7 +27,7 @@ class PG::Statement < ::DB::Statement
       raise "expected RowDescription or NoData, got #{frame}"
     end
     ResultSet.new(self, fields)
-  rescue IO::EOFError
+  rescue IO::Error
     raise DB::ConnectionLost.new(connection)
   end
 
@@ -38,7 +38,7 @@ class PG::Statement < ::DB::Statement
       rows_affected: result.rows_affected,
       last_insert_id: 0_i64 # postgres doesn't support this
     )
-  rescue IO::EOFError
+  rescue IO::Error
     raise DB::ConnectionLost.new(connection)
   end
 end


will:
This seems reasonable to me, but I haven't thought it through all the way. @bcardiff you added these rescues in #71 , does this change make sense to you?

omarroth:
> This seems reasonable to me, but I haven't thought it through all the way.

I'm in a similar position currently. It's fixed an issue in my application so I thought it would be worth raising here. Another pair of eyes would definitely be appreciated.

bcardiff:
The change seems safe, it will catch more situations than `IO::EOFError`.

I think the question is why a closed connection is been used? It's either a bug on crystal-db that should have it's own fix, or the app is holding a reference to a connection manually and been reused?

When a connection is closed some cleanup should be done to avoid reusing it https://github.com/crystal-lang/crystal-db/blob/be139c900c86e28b150b9808c74d2a4dd2bb913b/src/db/connection.cr#L59-L63 .

will:
@omarroth if you're able to, can you see if you can figure out where the connection-holding-on-to issue is coming from? If it is in crystal-pg, it'd be better to fix it here, but after a quick look nothing obvious is jumping out to me.

omarroth:
> can you see if you can figure out where the connection-holding-on-to issue is coming from?

Unfortunately I haven't been able to, hence this PR. 

crystal-lang/crystal-db#58 may be related, which was fixed with crystal-lang/crystal-mysql#35 (at least in crystal-mysql), but applying the relevant changes to this library did not resolve the issue I was encountering.

PR #175, opened 2019-07-24
Correctly encode timestamp values to consider microseconds

### LINKED ISSUE #140: Add support for TIMESTAMP columns in PQ::Param.encode
#### Background

* crystal-pg 0.14.1
* crystal-db 0.5.0
* PostgreSQL 9.6.8

When using parameterized queries, values in TIMESTAMP columns are not being matched correctly.

#### Example

```
# CREATE TABLE foo (
  ts TIMESTAMP NOT NULL DEFAULT NOW()
);
```

```
# INSERT INTO foo DEFAULT VALUES;
INSERT 0 1

# SELECT * FROM foo;
             ts
----------------------------
 2018-04-10 08:00:41.386075
(1 row)
```

But when we use Crystal, e.g.

```
rs = db.query_all("SELECT ts FROM foo;")
rs.move_next
ts = rs.read(Time)
p ts         # => 2018-04-10 08:00:41 UTC
p ts.class   # => Time
```

Now, if we try to `SELECT` or `DELETE` with a `WHERE` clause, we get no matches.

```
found = db.query_all("SELECT ts FROM foo WHERE ts = $1;", ts) do |rs|
  rs.read
end
p found        # => [] (no results)

exec_rs = db.exec("DELETE FROM foo WHERE ts = $1", ts)
p exec_rs.rows_affected # => 0_i64
```

#### Issue

`Time` values are encoded by `PQ::Param` using `to_s(ISO_8601)`. However, PostgreSQL expects `TIMESTAMP` values to have the form: `YYYY-MM-DD hh:mm:ss.nnnnnn`.

#### Workaround

The following workaround works:

```
# Encode the timestamp as a string in the format the PostgreSQL expects
pg_ts = ts.to_s("%Y-%m-%d %H:%M:%S.%6N")

found = db.query_all("SELECT ts FROM foo WHERE ts = $1;", pg_ts) do |rs|
  rs.read
end
p found # => [2018-04-10 08:00:41 UTC]

exec_rs = db.exec("DELETE FROM foo WHERE ts = $1", pg_ts)
p exec_rs.rows_affected # => 1_i64
```

asterite:
Fixes #140

I was able to reproduce the issue and fix suggested by @aisrael seems to work well.

I could extract the format string to a constant but I didn't know how to name it (maybe like this it's fine).

### DIFF
--- a/spec/pg/encoder_spec.cr
+++ b/spec/pg/encoder_spec.cr
@@ -25,6 +25,7 @@ describe PG::Driver, "encoder" do
 
   test_insert_and_read "timestamp", Time.utc(2015, 2, 3, 17, 15, nanosecond: 13_000_000)
   test_insert_and_read "timestamp", Time.utc(2015, 2, 3, 17, 15, 13, nanosecond: 11_000_000)
+  test_insert_and_read "timestamp", Time.utc(2015, 2, 3, 17, 15, 13, nanosecond: 123_456_000)
 
   test_insert_and_read "bool[]", [true, false, true]
 
--- a/src/pq/param.cr
+++ b/src/pq/param.cr
@@ -1,7 +1,6 @@
 require "../pg/geo"
 
 module PQ
-  ISO_8601 = "%FT%X.%L%z"
   # :nodoc:
   record Param, slice : Slice(UInt8), size : Int32, format : Int16 do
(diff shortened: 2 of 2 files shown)


will:
Awesome, thanks!

PR #176, opened 2019-07-25
Add scram support, switch to circle ci

### LINKED ISSUE #154: Works with scram-sha-256?
Subj.

will:
This doesn't yet do the channel binding, but that can be added later. Fixes #154


will:
I looked into the channel binding today, but I can't figure out how to get the peer cert from either the  `OpenSSL::SSL::Context::Client` or the `OpenSSL::SSL::Socket::Client`. I imagine new functions need to be mapped out of openssl, but I'm not sure where to start.

For channel binding to work, you need to need to 

> Then comes tls-server-end-point, which needs more routines from OpenSSL, first to get the server certificate and hash it:
>
> SSL_get_peer_certificate() to get the peer certificate data, which needs to be used on the client side in this case.
> Once this information is fetched, hashing it it first necessary. In this process comes a detail of RFC 5929: if the signature algorithm of a certificate is MD5 or SHA-1, then the hashing needs to be done with SHA-256. For anything above, use the same hashing. Getting a hash from a certificate is a matter of using X509_digest(), but the algorithm type to specify depends on the signature algorithm the certificate is using. Digging through the code of OpenSSL, I have found an answer of how to do that in crypto/asn1/a_verify.c:
> 
> Use X509_get_signature_nid() to find the signature algorithm, note that this may not be the hash itself.
> Then apply OBJ_find_sigid_algs() to find out the real algorithm.
from: https://paquier.xyz/postgresql-2/channel-binding-openssl/

libpq:
https://github.com/postgres/postgres/blob/a806b14a1139d3443a7c2c263abb9ff8e4a65694/src/interfaces/libpq/fe-secure-openssl.c#L374

@asterite do you know? Or I'm not sure who has done the most ssl/tls stuff with crystal

asterite:
Unfortunately I'm clueless regarding openssl (I think I didn't touch a single code of it in Crystal).

will:
Mayyyybe @ysbaddaden can you point me in the right direction?

straight-shoota:
There is currently no support to retrieve the certificate from `SSL::Context`. The implementation is pretty limited, it only allows to specify a file or director path where the certificates are to be found.

But it should probably not be too complicated to add `SSL_get_peer_certificate`. It receives the value from `OpenSSL::SSL::Context#@handle` and returns an X509 pointer. There is already a `OpenSSL::X509::Certificate` class, but it doesn't provide any access to the actual certificate data. So you'd have to use `X509_digest` directly.

will:
@straight-shoota sounds good, thanks!

src/pq/connection.cr
```diff
@@ -270,6 +271,69 @@ module PQ
...
+      def generate_client_final_message(body)
+        server_first_msg = String.new(body)
+        params = server_first_msg.split(',')
+        r = params.find { |p| p[0] == 'r' }.not_nil![2..-1]
+        s = params.find { |p| p[0] == 's' }.not_nil![2..-1]
+        i = params.find { |p| p[0] == 'i' }.not_nil![2..-1].to_i
+        raise ConnectionError.new("SASL: scram server nonce does not start with client nonce") unless r.starts_with?(@client_nonce)
+
+        client_final_msg_without_proof = "c=biws,r=#{r}"
+        salted_pass = OpenSSL::PKCS5.pbkdf2_hmac(@password, Base64.decode(s), i, algorithm: OpenSSL::Algorithm::SHA256, key_size: 32)
```
jkatz: @will The password needs to be normalized with SASLPrep (well, PostgreSQL flavored SASLPrep), otherwise some valid passwords will fail. Here is an example of how to do it:

https://github.com/MagicStack/asyncpg/blob/master/asyncpg/protocol/scram.pyx#L263

src/pq/connection.cr
```diff
@@ -270,6 +271,69 @@ module PQ
...
+      final_msg = ctx.generate_client_final_message(continue.body)
+
+      # send client-final-message
+      write_chr 'p'
+      write_i32 4 + final_msg.bytesize
+      soc << final_msg
+      soc.flush
+
+      # receive server-final-message
+      expect_frame Frame::Authentication
```
jkatz: @will After this you're supposed to verify the server signature, where you can confirm that the server actually knows the password, too (the server can lie to you and say "Yes, I know your password! Give me your secrets now").

Example for how to complete: https://github.com/MagicStack/asyncpg/blob/master/asyncpg/protocol/scram.pyx#L176

src/pq/connection.cr
will: On https://www.postgresql.org/docs/current/sasl-authentication.html#SASL-SCRAM-SHA-256 it says that scram requires utf8 and postgres doesn't, which is the reason to do this saslprep. However, all `String`s in Crystal are UTF8, and there is no way to use a not-`String` to crystal-pg as it is. So maybe there is no reason to do this until (and if) non-string passwords are allowed?

jkatz: @will So what PostgreSQL does is that any UTF8 string goes through SASLprep. If it's not a UTF-8 string, or if it fails at certain parts of the SASLprep, then it just passes the string through. The server-side implementation can be found here:

https://git.postgresql.org/gitweb/?p=postgresql.git;a=blob;f=src/common/saslprep.c

Certain password will fail without following it. Here's an example of some of the test cases to try:

https://github.com/MagicStack/asyncpg/blob/master/tests/test_connect.py#L238

will: moved this to https://github.com/will/crystal-pg/issues/181

src/pq/connection.cr
will: This has been fixed

jkatz: @will Awesome!

Neustradamus:
Linked to:
- https://github.com/scram-sasl/info/issues/1

PR #183, opened 2019-08-08
Fix encoding format of time with no-zero offset

### LINKED ISSUE #182: When time is inserted into postgres, 4 hours get added to it.
I recently updated from crystal-db 0.5.1 and crystal-pg 0.16.1 to crystal-db 0.6.0 and crystal-pg 0.18. This caused the following change in the application behavior, which appears to be a bug.

Behavior with the newer versions of these packages
(1) I send an integer = 1565713800000 milliseconds since epoch to the server as part of JSON. This equals 12:30pm US Eastern time on August 13.
(2) On server I use timecreated: {type: Time?, converter: Time::EpochMillisConverter} as part of JSON.mapping to read this value.
(3) I insert the read value (which is now of type Time) into Postgres timestamptz column. The time in postgres shows up as 4:30pm US Eastern time. Postgres may be saving the timestamp, but in psql it displays the time locally (which is US Eastern time for me).
(4) As expected, when I get this value out again on another page, it is now 4 hours ahead of what I expect.

I reverted crystal-pg to 0.16.1 (which automatically changed crystal-db to 0.5.1). The behavior is as expected. When I send the integer = 1565713800000 milliseconds, and follow the above steps, the time in Postgres shows as 12:30pm US Eastern time on August 13.

straight-shoota:
Fixes #182

### DIFF
--- a/spec/pg/encoder_spec.cr
+++ b/spec/pg/encoder_spec.cr
@@ -26,6 +26,7 @@ describe PG::Driver, "encoder" do
   test_insert_and_read "timestamp", Time.utc(2015, 2, 3, 17, 15, nanosecond: 13_000_000)
   test_insert_and_read "timestamp", Time.utc(2015, 2, 3, 17, 15, 13, nanosecond: 11_000_000)
   test_insert_and_read "timestamp", Time.utc(2015, 2, 3, 17, 15, 13, nanosecond: 123_456_000)
+  test_insert_and_read "timestamptz", Time.local(2019, 8, 13, 12, 30, location: Time::Location.fixed(-14_400))
 
   test_insert_and_read "bool[]", [true, false, true]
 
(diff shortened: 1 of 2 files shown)


bcardiff:
I believe RFC_3339 will send the nanoseconds instead of microseconds as #175 intended. I am unaware if that is an actual issue.

straight-shoota:
Yes, it includes 9 sub-second digits. But postgres can understand that, so it's not really an issue.

There is a `fraction_digits` argument to `RFC_3339.format` but it's currently not implemented to function as a size limit. We could probably make that work and get the intended microsecond output.

straight-shoota:
Btw. Golang driver seems to send 9 digits as well https://github.com/lib/pq/blob/master/encode.go#L487

will:
released v0.18.1 with just this fix

will:
Thank you :)

PR #184, opened 2019-08-09
Fix ipv6 hostname connection

j8r:


src/pq/conninfo.cr
```diff
@@ -36,7 +36,7 @@ module PQ
     def initialize(host : String? = nil, database : String? = nil, user : String? = nil, @password : String? = nil, port : Int | String? = 5432, sslmode : String | Symbol? = nil)
       @host = default_host host
       db = default_database database
-      @database = db.starts_with?('/') ? db[1..-1] : db
```
straight-shoota: Why this change? It doesn't seem to be related to IPv6 hostname and changes the behaviour.

j8r: Yes that's lchop, a mistake.
That's in an other commit.

straight-shoota: `db.lchop?('/') || db` == `db.chop('/')`

will:
Can you please explain what the fix is? Like an example of how it was going wrong and now right? Is it testable to prevent it from breaking in the future?

j8r:
It was @MrSorcus that reported the issue initially.
There was a `No address found for [::]:5432 over TCP (Socket::Addrinfo::Error)` when using an IPv6 address like `DB.open("postgres://test@[::]:5432/test_db")`.
Now, it works with the patch.

straight-shoota:
crystal-lang/crystal#8066 fixes the same bug with `HTTP::Websocket`.

A spec would be great here, too!

asterite:
I would keep optimizations and other stylistic changes outside of this PR, though, they make it harder to review (and some. changes I don't agree, and will might not agree either)

j8r:
@asterite not if you review per commit, there are very few changes so I think it's OK.
For the spec, sure I add it one right now.

will:
Thanks! I rebased and merged this, but outside of github so it looks like it can't figure it out.

PR #185, opened 2019-09-20
Update to crystal-db ~> 0.7.0

bcardiff:
There is a new release of crystal-db with some breaking changes.

Ref: https://github.com/crystal-lang/crystal-db/pull/110

### DIFF
--- a/shard.yml
+++ b/shard.yml
@@ -4,4 +4,4 @@ version: 0.18.1
 dependencies:
   db:
     github: crystal-lang/crystal-db
-    version: ~> 0.6.0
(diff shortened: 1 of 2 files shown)


bcardiff:
If we can have a crystal-pg release as 0.19.0 in the following days it would be ideal for the web frameworks to upgrade. 🙇

This should be compatible with the upcoming Crystal 0.31.0

will:
Cool, released v0.19.0 just now

bcardiff:
Thanks a lot! 🙇

Issue #164, opened 2019-04-13
Support range types


vladfaust:
Currently:

```crystal
require "pg"

db = DB.open("postgres://postgres:postgres@localhost:5432/postgres")
pp db.scalar("SELECT int4range(10, 20)") # => Bytes
```

miketheman:
I've been taking a stab at this, and can't seem to figure out where the "rules" for decoding a given message from Postgres wire format come from - @will any advice here?

asterite:
@miketheman It's in https://github.com/will/crystal-pg/blob/master/src/pg/decoder.cr

Every postgres type has an ID, called `oid`. A decoder can handle a series of `oid`s. So my guess is that you'll need to define a new `Int4Range` decoder and register it, but then you'll also want to define a `Postgres::Range` type, which is different than Crystal's `Range`.

(this is just from me fiddling with the library, I might be wrong)

miketheman:
@asterite Thanks - I got that far. I'm more interested in how to interpret the bytes returned from PG.

will:
The functions in postgres are usually ending in `_send`  

Here is the range one I think:  https://github.com/postgres/postgres/blob/b538c90b1bded5464787e2b8e4431302d24eb601/src/backend/utils/adt/rangetypes.c#L246

will:
You can also call them as functions in postgres if that helps playing around with things faster:

```
will=# select range_send('[1,3]'::int4range);
              range_send
--------------------------------------
 \x0200000004000000010000000400000004
(1 row)
```

miketheman:
I started a Draft PR for discussion.

pascalbetz:
I can work on this, if it is stalled.

PR #190, opened 2019-10-09
feat: support int4range type

miketheman:
This is an exploratory effort to support #164

It is not ready for merging, but I wanted to to open it for discussion
and review.

### DIFF
--- a/spec/pg/decoders/range_decoder_spec.cr
+++ b/spec/pg/decoders/range_decoder_spec.cr
@@ -1,0 +1,10 @@
+require "../../spec_helper"
+
+describe PG::Decoders do
+  describe "int4range" do
+    test_decode "exclusive", "'[10,20)'::int4range", Range.new(10, 20)
+    test_decode "exclusive", "'(10,20]'::int4range", Range.new(11, 21)
+    test_decode "inclusive", "'[10,20]'::int4range", Range.new(10, 21)
+    test_decode "inclusive", "'[10,20)'::int4range", Range.new(10, 20)
+  end
+end
--- a/spec/pg/decoders/range_decoder_spec.cr
+++ b/spec/pg/decoders/range_decoder_spec.cr
@@ -6,5 +6,6 @@ describe PG::Decoders do
     test_decode "exclusive", "'(10,20]'::int4range", Range.new(11, 21)
     test_decode "inclusive", "'[10,20]'::int4range", Range.new(10, 21)
     test_decode "inclusive", "'[10,20)'::int4range", Range.new(10, 20)
+    test_decode "negatives", "'[-14,-5)'::int4range", Range.new(-14, -5)
   end
 end
--- a/src/pg/decoder.cr
+++ b/src/pg/decoder.cr
@@ -471,6 +471,44 @@ module PG
       end
     end
 
+    struct RangeDecoder
+      include Decoder
+
+      def_oids [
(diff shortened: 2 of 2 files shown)


src/pg/decoder.cr
```diff
@@ -471,6 +471,44 @@ module PG
...
+
+        # TODO: add `exclusive` flag? Do we need to, or since the decoder
+        # is already receiving a discrete range type, maybe we don't?
+        # Ref https://www.postgresql.org/docs/10/rangetypes.html#RANGETYPES-DISCRETE
+        Range.new(lower_bound, upper_bound)
+      end
+
+      def type
+        # TODO: do I actually need a new PG-specific type, or can we use real Crystal Ranges?
+        Range
```
miketheman: Should this be a specifically named `Int4Range`, or should the Range decoder emit a Crystal Range? Is the expectation that the end user desire to be super-specific about whether they are passing an Integer, BigInt, or Numeric type?

spec/pg/decoders/range_decoder_spec.cr
```diff
@@ -0,0 +1,10 @@
+require "../../spec_helper"
```
miketheman: I created a new spec file to help organize the range specs outside of the regular specs - it looked like that's what happened with `array_decoder_spec.cr`
I guess if this turns into multiple structs for the different ranges, then maybe the decoder ought to be extracted as well?

spec/pg/decoders/range_decoder_spec.cr
```diff
@@ -0,0 +1,10 @@
...
+
+describe PG::Decoders do
+  describe "int4range" do
+    test_decode "exclusive", "'[10,20)'::int4range", Range.new(10, 20)
+    test_decode "exclusive", "'(10,20]'::int4range", Range.new(11, 21)
+    test_decode "inclusive", "'[10,20]'::int4range", Range.new(10, 21)
+    test_decode "inclusive", "'[10,20)'::int4range", Range.new(10, 20)
+  end
```
will: There was recently a regression on a different part because we weren't testing negative numbers, so I think we should make sure to include at least one test with something like '[-14,-5]'

spec/pg/decoders/range_decoder_spec.cr
will: Yeah I don’t have any fully thought out reason, but it kinda makes sense since there will be a lot of related decoders for the ranges.

src/pg/decoder.cr
will: It's been a while since I've looked into ranges in crystal, but I think maybe postgres ranges have more "features" than crystal ranges. 

In some cases, it is probably easy to normalize a postgres range into a crystal one, like how in the tests you have here the inclusive/exclusive adding 1. However in other cases it might not work, like for the float ranges and date ranges allowing positive and negative infinity. We may have to make our own range class for those. Or enhance the crystal stdlib ranges.

spec/pg/decoders/range_decoder_spec.cr
miketheman: Good call, I've added that test.

src/pg/decoder.cr
straight-shoota: `Range` can take `nil` as start and/or end value, so maybe this could be used to represent infinity.

PR #199, opened 2020-01-04
Update shard.yml

grkek:


will:
Sorry, but I think this would break the shard. Can you confirm @bcardiff?

bcardiff:
Yes. The db drivers use a bit more conservative dependencies to ensure a valid crystal-db is used.

@grkek I guess you have an application that has a dependency of `crystal-db ~> 0.7.0` and that is causing a conflict when trying to bump to the latest `crystal-pg`. Depending on you app/shard the solution could be either

* remove crystal-db as an explicit dependency
* update the version in your shard.yml
* apply these on some other nested dependency

grkek:
> Yes. The db drivers use a bit more conservative dependencies to ensure a valid crystal-db is used.
> 
> @grkek I guess you have an application that has a dependency of `crystal-db ~> 0.7.0` and that is causing a conflict when trying to bump to the latest `crystal-pg`. Depending on you app/shard the solution could be either
> 
>     * remove crystal-db as an explicit dependency
> 
>     * update the version in your shard.yml
> 
>     * apply these on some other nested dependency

I had conflicts with pg and crecto shards, whenever I tried to install it threw an error then, I removed the shard.lock and it fixed itself somehow even tho I did the very same thing before and it didn't work ?, cheers.

PR #203, opened 2020-04-06
Crystal 0.34.0

straight-shoota:
This PR updates for compatibility with newly release Crystal 0.34.0 addressing the breaking changes from crystal-lang/crystal#8885 and crystal-lang/crystal#8424.

PR #204, opened 2020-04-07
Update to crystal-db ~> 0.9.0

bcardiff:
And keep support for older than Crystal 0.34.0 https://github.com/will/crystal-pg/pull/203#issuecomment-610082732. 

Some tools state that they work with also previous versions of Crystal.

Since we are doing some heavy cleanup for 1.0, soon we can remove them. I usually encourage at least previous + latest support to simplify the migration of apps codebases.

If a v0.21.0 can be released as is it would be great. Let me know if I can give you a hand with something else @will .

### DIFF
--- a/shard.yml
+++ b/shard.yml
@@ -4,4 +4,4 @@ version: 0.20.0
 dependencies:
   db:
     github: crystal-lang/crystal-db
-    version: ~> 0.8.0
+    version: ~> 0.9.0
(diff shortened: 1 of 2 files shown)


bcardiff:
If you find it handy, here is a changelog. 🙈 

```
v0.21.0    2020-04-??
=====================
* Support Crystal v0.34.0 (thanks @straight-shoota, @bcardiff)
* Update crystal db support to 0.9.0 (thanks @bcardiff)
* Add Error Handling sample in README (thanks @ArtLinkov)
```

will:
Thanks for the changelog! :)

PR #206, opened 2020-05-22
Upgrade support to Crystal 0.35

bcardiff:
WIP

### DIFF
--- a/src/pq/connection.cr
+++ b/src/pq/connection.cr
@@ -319,7 +319,11 @@ module PQ
       end
 
       private def sha256(key)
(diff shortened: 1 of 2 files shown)


bcardiff:
@will this is good to go and is backward compatible.

Since the only change from v0.21.0 is the addition of Crystal 0.35.0 support without big changes I think v0.21.1 can do. But as you prefer.

There is no need for a release of crystal-db.

In the following crystal releases we will need to add a `crystal:` property  in the shard.yml file. For now it's not needed. There were some changes in shards 0.11 regarding dependency resolution.

will:
Thanks! Looking forward to crystal 1.0 :)

PR #210, opened 2020-09-14
Update crystal-db for logging support

bcardiff:
Depends on https://github.com/crystal-lang/crystal-db/pull/134

### DIFF
--- a/shard.yml
+++ b/shard.yml
@@ -4,4 +4,6 @@ version: 0.21.1
 dependencies:
   db:
     github: crystal-lang/crystal-db
-    version: ~> 0.9.0
+    version: ~> 0.10.0
+
+crystal: 0.35.0
--- a/src/pg/statement.cr
+++ b/src/pg/statement.cr
@@ -1,6 +1,6 @@
 class PG::Statement < ::DB::Statement
-  def initialize(connection, @sql : String)
-    super(connection)
+  def initialize(connection, command : String)
+    super(connection, command)
   end
 
   protected def conn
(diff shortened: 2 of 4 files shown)


will (approved):
Nice!

I've really been liking the new stdlib log system, btw

jwoertink:
Now that crystal-db 0.10 is out, this will be great to get out there! 🎉

bcardiff:
Since crystal-db ~> 0.10.0 requires Crystal: 0.35.0 there were some clean-up that could be done.

I added the `crystal: 0.35.0` to the shard.yml since it is almost mandatory for a shard to include that information. The implicit value is `< 1.0.0`.

will:
@bcardiff I'm a little unclear at all the details here. While it doesn't seem complicated per se, it feels like there are subtle details with all the different projects I don’t want to screw up. 

I'm unavailable for the next few days, so feel free to merge whatever is correct if needed.

bcardiff:
Usually crystal-db is conservative with respect the minimum crystal std-lib needed. With the integration of the recent logging module that is a promise harder to keep. crystal-db 0.10.0 requires crystal: 0.35.0. So there is no addition restriction that this PR is adding.

Existing projects, as long as they use a shard.lock or a crystal-pg conservative restriction, should be fine. They will use prior versions of the dependencies.

To make it available a v0.22.0 version should be released. With at least this PR been the only change from v0.21.1. There is no rush on my end. But if you prefer I can do the release commit. Your call.

---

On a note to frameworks ;-), crystal-db 0.9 and 0.10 are compatible as a consumer. So if the framework used wil/crystal-pg ~> 0.21.0 as a restriction, it would require at least a patch release of the framework with updated crystal-pg dependency. Either allowing 0.21 and 0.22 `>= 0.21.0, <= 0.23.0`, or only `~> 0.22.0`. It's up to the framework choice that.

will:
> To make it available a v0.22.0 version should be released. With at least this PR been the only change from v0.21.1. There is no rush on my end. But if you prefer I can do the release commit. Your call.

If not having the release for a couple of days is causing problems for people, please go ahead :) Otherwise I can take care of it then.

will:
Thanks! Just released `0.22.0` with this.

Issue #211, opened 2020-09-30
Support for Interval


pascalbetz:
Support for INTERVAL type:

https://www.postgresql.org/docs/current/datatype-datetime.html

Currently Intervals are decoded as Bytea.
I can take a stab at it but - as I just started with Crystal - I'm unsure about the target Crystal type. Time::Span does not cover the full range of INTERVAL. Is there anything else that this could be mapped to?

pascalbetz:
For reference: https://github.com/postgres/postgres/blob/a094c8ff53523e88ff9dd28ad467618039e27b58/src/backend/utils/adt/timestamp.c#L1003

I am currently thinkering around and my solution converts INTERVAL to Time::Span but raises when Month/Year are present.

will:
Hi, thanks for opening this issue :)

You're right that there is no full-fidelity mapping of postgres `interval` type to a crystal std lib type. I think you understand why but just going to copy the docs here for other people coming to this issue who perhaps haven't worked with postgres `interval`s much.

> Internally interval values are stored as months, days, and seconds. This is done because the number of days in a month varies, and a day can have 23 or 25 hours if a daylight savings time adjustment is involved. The months and days fields are integers while the seconds field can store fractions. Because intervals are usually created from constant strings or timestamp subtraction, this storage method works well in most cases, but can cause unexpected results:

```
SELECT EXTRACT(hours from '80 minutes'::interval);
 date_part
-----------
         1

SELECT EXTRACT(days from '80 hours'::interval);
 date_part
-----------
         0
```
> Functions justify_days and justify_hours are available for adjusting days and hours that overflow their normal ranges.

As for solutions here are a couple of ways, though there are probably others.
1. as you suggest, do what we can to coerce safely into the crystal type and error when we cant
2. always to coerce to the crystal type by pretending that all months are 30 days, etc
3. create our own `struct PG::Interval` that behaves identically to the postgres type

1 is good, however I don’t like the possibility of a query that otherwise is perfectly fine for postgres to throw an exception from the driver just because it includes a field that has an interval with months. I think there are enough cases where programs don't actually look at all of the data returned, so this is a little too opinionated for a default I think.

I don’t like 2, but just wanted to enumerate it.

While it's more work I think 3 makes the most sense with the general spirit of postgres to not lose precision if it's there. This struct could have a method called something like `#to_span` that does like (1) and converts or raises if it can't.

[...]

PR #212, opened 2020-09-30
Add support for Postgres INTERVAL

pascalbetz:
https://github.com/will/crystal-pg/issues/211

Add support for INTERVAL.

I just started learning Crystal (which so far works well since I'm coming from Ruby) so any feedback is welcome. 

`PG::Interval` allows to convert to `Time::Span`and `Time::MonthSpan` but does not do any roll-over between those spans (does not add days to months).

Some questions:
1) What exception to raise when `PG::Interval` can not be contained in `Time::Span`?
2) Naming of the conversion methods? `to_span` / `to_month_span` is there anything better?
3) Is `Interval#to_spans` useful at all?

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,6 +133,7 @@ Since it uses protocol version 3, older versions probably also work but are not 
 - regtype
 - geo types: point, box, path, lseg, polygon, circle, line
 - array types: int8, int4, int2, float8, float4, bool, text, numeric, timestamptz, date, timestamp
+- interval (2)
 
 1: A note on numeric: In Postgres this type has arbitrary precision. In this
     driver, it is represented as a `PG::Numeric` which retains all precision, but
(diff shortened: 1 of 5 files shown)


will:
Thank you for this. I squashed your commits and rebased them (outside of github so it cant tell the pr has been accepted), and made a couple of changes myself. I think just the default runtime error is good, as is the method names.

I changed around the `ignore_overflow` idea to let people optionally approximate months to a number of days of their choosing rather than just drop the months altogether. Raising by default is good so it's not losing precision, but this allows people to either self opt-in to something like `30` and get a probably good enough approximation, or give `0` and have the behavior you had in the patch where it just drops months altogether.

I think maybe some future work could add comparison methods, but it's probably best to get a feel for what is needed from people who are using this for real, so that can wait until then.

I'll take a look at your other patch soon :)

PR #213, opened 2020-10-03
add support for ranges

pascalbetz:
Implements Ranges
(https://github.com/will/crystal-pg/issues/164)

This is as far as i can get with my limited Crystal knowhow...so i'll needs some help.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,6 +133,7 @@ Since it uses protocol version 3, older versions probably also work but are not 
 - regtype
 - geo types: point, box, path, lseg, polygon, circle, line
 - array types: int8, int4, int2, float8, float4, bool, text, numeric, timestamptz, date, timestamp
+- range: int4range, int8range, daterange, tsrange, tstzrange, numrange
 
 1: A note on numeric: In Postgres this type has arbitrary precision. In this
     driver, it is represented as a `PG::Numeric` which retains all precision, but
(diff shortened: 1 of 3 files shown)


will:
Thanks for putting this together. I'm not available for a couple of days, but I'll take a look at this next week and give feedback then :)

pascalbetz:
**Begin/Endless Ranges**

Postgres supports "infinite" values for lower/upper boundary. This corresponds to begin/endless Ranges in Crystal. 
E.g.: `[, 50)` => `nil..50`
This makes the Range that gets created `Range(T | Nil, T | Nil)`
Is this OK?

Another possibility would be to use the Datatypes MIN/MAX value (e.g. Int64::MIN) and make sure the boundary is inclusive:
E.g.: `[, 50)` => `Int32::MIN..50` so we would not create begin/endless Ranges. (This would not map what is in Postgres though, so I don't like it)

**Empty Ranges**

Postgres empty ranges are currently implemented as exclusive Ranges with the same value for lower and upper boundary.
E.g. `0...0`

The values I picked to build these empty ranges are arbitrary though and this does not map exactly what is stored in Postgres.

There is no way to differentiate between 

`(,)::int4range` and `[0,0]::int4range` because both values end up in `Range.new(0, 0)`.

What do you think?

pascalbetz:
@will
 I plan on fixing the build and adding multiranges (Postgres 14) support if this can get merged. Let me know what you think.

PR #219, opened 2020-12-17
[CI] Run check_format as separate job

straight-shoota:
It's annoying when CI fails early because of a format error and doesn't run specs.
Also formatter is independent of postgres version matrix.


will:
Nice, thanks!

PR #220, opened 2020-12-17
Add config setting auth_methods

straight-shoota:
This adds a `auth_methods` config setting which allows to configure which auth methods the client should accept.

The settings value is a list of comma separated values. The default is `scram-sha-256,md5`. `cleartext` needs to be explicitly enabled.

Is there any place where documentation can be added?

I didn't do anything about `OK` authentication frame. A safe default should also fail if the server just passes you through without a credential check.
But this is more complicated because it is fine with SSL client certificate for example.
So it needs more work. For now I'm just posting this patch.

Superseeds #218
cf crystal-lang/crystal-db#141

### DIFF
--- a/src/pq/conninfo.cr
+++ b/src/pq/conninfo.cr
@@ -5,6 +5,8 @@ module PQ
   struct ConnInfo
     SOCKET_SEARCH = %w(/run/postgresql/.s.PGSQL.5432 /tmp/.s.PGSQL.5432 /var/run/postgresql/.s.PGSQL.5432)
 
+    SUPPORTED_AUTH_METHODS = ["cleartext", "md5", "scram-sha-256"]
+
     # The host. If starts with a / it is assumed to be a local Unix socket.
     getter host : String
 
@@ -31,6 +33,8 @@ module PQ
 
     # The sslrootcert. Optional.
     getter sslrootcert : String?
+
+    getter auth_methods : Array(String) = ["scram-sha-256", "md5"] of String
 
     # Create a new ConnInfo from all parts
     def initialize(host : String? = nil, database : String? = nil, user : String? = nil, @password : String? = nil, port : Int | String? = 5432, sslmode : String | Symbol? = nil)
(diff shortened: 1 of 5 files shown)


src/pq/conninfo.cr
```diff
@@ -32,6 +34,8 @@ module PQ
     # The sslrootcert. Optional.
     getter sslrootcert : String?
 
+    getter auth_methods : Array(String) = ["scram-sha-256", "md5"] of String
```
will, line 37: Should we maybe use symbols or some other not-string type here? It probably doesn't make huge difference, but it feels like using strings in crystal when you're not manipulating them is something Asterite would frown at.

will (commented):
Overall this is good, and meets what I was looking for in having cleartext be opt-in. Thank you

> Is there any place where documentation can be added?

The only place right now is the ever growing and not well organized readme. I suppose we should eventually build some better doc thing sooner or later.

> anything about OK authentication frame. 

Yeah this is a little more complex. For the reason you say and also, if I remember correctly, if you're connecting over like a local unix socket, and pg is doing ident/peer auth, you just get `OK` and that is legitimately ok. I think we can probably leave this one be for now, but we can wait a bit to see if there are any counter examples. 

But again, overall looks good :) Let me know when you think it's ready ready.

src/pq/conninfo.cr
will: I know it'll come in as strings on the url parsing side, but maybe not allocating these 2 strings every time adds up. Not sure

straight-shoota: No, strings are totally the best choice here. We need to match against the configuration data which is... a string. So if this were symbols, they would need to be converted to string in order to compare it to a string. 0 benefit from symbols.
Actually, Asterite has repeatedly said that he would rather not have symbols in the language. So has Matz btw. regarding Ruby ;) Some people like symbols, but i think it's mostly a style thing. My wish for Crystal 2.0 (or later) is to remove them 😆

straight-shoota:
I think this is ready to merge. As you said, further improvements can follow later.

src/pq/conninfo.cr
will: Alright, makes sense :)

will:
> I think this is ready to merge. As you said, further improvements can follow later.

Sounds good, I'll do this tomorrow.

will:
Thank you! Going to wrap up a few other things then do a release soon.

PR #221, opened 2020-12-31
Read Slice(UInt8) as String if given as type

### LINKED ISSUE #43: CITEXT type is understood as Slice instead of String in 0.7.1
In version `cafe4748b64a9c95ab93b4c19315780e8e254291`, querying a [CITEXT](https://www.postgresql.org/docs/9.4/static/citext.html) type field would return the result as String, so the following returned the expected result:

``` crystal
DB.exec({String}, "SELECT name FROM users").rows[0][0]
```

As of `0.7.1`, the code above started throwing `cast to String failed (TypeCastError)`, though the following did work and returned the same value:

``` crystal
String.new(DB.exec({Slice}, "SELECT name FROM users").rows[0][0])
```

It seems that a change in the decoder caused CITEXT fields to stop being understood as actual Strings.

Tested with postgresql `9.4.7`.

matthewmcgarvey:
Fixes #43

In order to support the citext extension in Avram, we need to be able to treat the column as a String. Two solutions were laid out in the connected issue:

- A: Calling `rs.read(String)` on a string-like column (citext, enum or similar), it should return a String.
- B: Calling `rs.read` on a string-like column, it should probably also return a String.

B was considered the more "complete solution" but also the more complicated. Since it's been 4 years since the issue was opened and over a year since the potential solutions were laid out, I went with the more immediate option A. If the read type is `Slice(UInt8)`, which is the type of a citext column, and `#read` was called with a type of `String`, then convert it to String. This helps with citext interop and I'm sure it helps with other column types as well.

### DIFF
--- a/spec/pg/driver_spec.cr
+++ b/spec/pg/driver_spec.cr
@@ -95,6 +95,19 @@ describe PG::Driver do
     end
   end
 
+  it "treats Slice(UInt8) as String" do
+    with_db do |db|
+      db.exec "create extension if not exists citext with schema public"
+      db.exec "drop table if exists users"
+      db.exec "create table users (email citext)"
+      db.exec "insert into users values ($1)", "foo@example.com"
+
+      db.query "select email from users limit 1" do |rs|
+        assert_single_read rs, String, "foo@example.com"
+      end
+    end
+  end
+
   describe "transactions" do
     it "can read inside transaction and rollback after" do
       with_db do |db|
--- a/spec/pg/driver_spec.cr
+++ b/spec/pg/driver_spec.cr
@@ -105,6 +105,10 @@ describe PG::Driver do
       db.query "select email from users limit 1" do |rs|
         assert_single_read rs, String, "foo@example.com"
       end
+
+      db.query "select email from users limit 1" do |rs|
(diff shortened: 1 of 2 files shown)


matthewmcgarvey:
This is related to https://github.com/will/crystal-pg/pull/89 which is attempting a more complete implementation laid out here https://github.com/will/crystal-pg/issues/43#issuecomment-221906216 but stalled out.

will:
Thanks! This looks good to me. I don’t myself use citext though, so I'm going to leave this open for a couple of days in case anyone has comments. But I expect to merge this, then do a release sometime early next week.

straight-shoota (approved):

jwoertink (approved):

bcardiff (approved):
At some point conversions like this should be handled in crystal-db directly as long as they are type-safe and without configuration. But to this some changes in the design need to be made. So, this is they for now.

bcardiff:
Oh! I remember. Double check that when using `rs.read(String?)` the conversion will work. It will probably wont. In crystal-sqlite3 there are a couple of these IIRC.

will:
Thanks!

PR #226, opened 2021-01-26
Update for Crystal 0.36 compatibility

bcardiff:
This PR makes the spec build on Crystal 0.36 without warnings.
The changes are backward compatibly with 0.35.

Another optional change is to declare this shard compatible with 1.0.0-x which should be almost 0.36. Is not mandatory and --ignore-crystal-version can be used otherwise.

### DIFF
--- a/shard.yml
+++ b/shard.yml
@@ -6,4 +6,4 @@ dependencies:
     github: crystal-lang/crystal-db
     version: ~> 0.10.0
 
-crystal: 0.35.0
+crystal: ">= 0.35.0, < 2.0.0"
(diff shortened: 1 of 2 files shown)


will (approved):
I don’t have all of the details of versioning in my head enough at the moment to be able to make a call the best path, so I'll trust you. Let me know what path you decide to go this way in the pr or the other way you mention in the comment. 

Also does this require a new small version bump?

will:
Also semi-related please take a look at https://github.com/crystal-lang/crystal/pull/8005 in case that's fallen off your radar. It would be nice to get that in so that this driver could support scram with channel binding.

bcardiff:
There is little need for a release right now. The latest release works fine for 0.36.
But a patch release would not hurt.

To get up to date with version things I can point to https://crystal-lang.org/2020/08/20/preparing-our-shards-for-crystal-1.0.html

Sorry crystal-lang/crystal#8005 was totally out my radar until now. I'll try to give feedback.

PR #232, opened 2021-07-05
Json columns are read as JSON::PullParser instead of JSON::Any

### LINKED ISSUE #125: Premature parsing of json
To parse json into my own type using `JSON.mapping`, I have to do:
```crstal
  def self.from_rs(rs : DB::ResultSet)
    from_json(rs.read(JSON::Any).to_json)
  end
```
This marshal/unmarshal thrashing could be avoided if `pg` just returned `JSON::PullParser` or some other type that doesn't scrap the ability to marshal via `JSON.mapping`

matthewmcgarvey:
Fixes #125 

Currently, if you have a json column and read it with this library, it is automatically turned into a `JSON::Any`. It makes sense when you don't care about the structure, but there's two problems with doing that:
1. `JSON::Any` is immutable so that values cannot be changed and then saved back to the database
2. You cannot convert a `JSON::Any` into a `JSON::Serializable`

This PR changes the default deserialization of json columns from `JSON::Any` to `JSON::PullParser`. The benefit we get from this is that it can easily be converted into a `JSON::Any` or a `JSON::Serializable`. It is still not mutable but it's easier to get to a mutable setting:

```crystal
properties = result_set.read(JSON::PullParser)
mutable_properties = Hash(String, String).new(properties)
```

I added a method specifically for still supporting calling `result_set.read(JSON::Any)` with hopes that will cause the least amount of breakage, because this is definitely a breaking change.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,6 @@
 v?.?.?     upcoming
 =====================
+* Decode json columns as JSON::PullParser instead of JSON::Any (thanks @matthewmcgarvey)
 
 v0.23.2    2020-03-27
 =====================
--- a/spec/pg/decoder_spec.cr
+++ b/spec/pg/decoder_spec.cr
@@ -38,11 +38,11 @@ describe PG::Decoders do
     UUID.new("7d61d548-124c-4b38-bc05-cfbb88cfd1d1")
 
   if Helper.db_version_gte(9, 2)
-    test_decode "json", %('[1,"a",true]'::json), JSON.parse(%([1,"a",true]))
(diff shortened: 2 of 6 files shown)


bcardiff (approved):
LGTM. I feel we need a read_io directly that will allow us to use the stream directly into de pullparser or any other consumer of IOs. I'm mentioning this to share this idea that should be done in crystal-db at some point I think.

will:
Thanks! Could you please add a line to the changelog file about this?

will:
Awesome, thanks!

PR #234, opened 2021-07-21
Add scram channel binding (scram-sha-265-plus)

will:
This will be preferred when the Postgres server offers it, as it is
strictly better than scram without channel binding. This was a new
feature in Postgres 11. Postgres 10 supports scram, but not channel
binding.

It requires a few extra openssl functions linked, which did not become a
part of Crystal until version 1.1.0, so I've backported that patch for
earlier versions of Crystal.

### DIFF
--- a/src/ext/openssl.cr
+++ b/src/ext/openssl.cr
@@ -1,0 +1,84 @@
+require "openssl"
+
+class OpenSSL::X509::Certificate
+  def scram_signature
+    # The TLS server's certificate bytes need to be hashed with SHA-256 if
+    # its signature algorithm is MD5 or SHA-1 as per RFC 5929
+    # (https://tools.ietf.org/html/rfc5929#section-4.1).  If something else
+    # is used, the same hash as the signature algorithm is used.
+    algo_type = signature_algorithm
+    algo_type = "SHA256" if algo_type == "MD5" || algo_type == "SHA1"
+    digest algo_type
+  end
+end
+
+# Backport of https://github.com/crystal-lang/crystal/pull/8005
+# for Crystal versions < 1.1.0. Can be removed once crystal-pg no longer
+# supports those versions of Crystal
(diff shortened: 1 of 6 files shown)


src/ext/openssl.cr
```diff
@@ -0,0 +1,84 @@
+require "openssl"
+
+class OpenSSL::X509::Certificate
+  def scram_signature
```
bcardiff, line 4: Is this something scoped to PG or is something that would make sense to port into the std-lib also?

src/pq/connection.cr
```diff
@@ -304,7 +320,14 @@ module PQ
         i = params.find { |p| p[0] == 'i' }.not_nil![2..-1].to_i
         raise ConnectionError.new("SASL: scram server nonce does not start with client nonce") unless r.starts_with?(@client_nonce)
 
-        client_final_msg_without_proof = "c=biws,r=#{r}"
+        if @signature
+          # cd10b...Cws == base64 of "p=tls-server-end-point,,"
+          b64sig = Base64.strict_encode @signature.not_nil!
```
straight-shoota, line 325:
```suggestion
        if signature = @signature
          # cd10b...Cws == base64 of "p=tls-server-end-point,,"
          b64sig = Base64.strict_encode signature
```

src/pq/connection.cr
```diff
@@ -304,7 +320,14 @@ module PQ
...
+          client_final_msg_without_proof = "c=cD10bHMtc2VydmVyLWVuZC1wb2ludCws#{b64sig},r=#{r}"
```
straight-shoota, line 326: Why not include encoding the static part (`p=tls-server-end-point,,`)? I think this makes the code much easier to read.
Performance impact should be completely negligible.

Neustradamus:
Linked to:
- https://github.com/scram-sasl/info/issues/1

Issue #195, opened 2019-11-07
Error handling for connect_listen


straight-shoota:
When closing the connection returned by `connect_listen`, an exception is raised inside the listen fiber:

```cr
require "pg"
connection = PG.connect_listen ENV["DATABASE_URL"], "foo" {  }
sleep 1
connection.close
```
```
Unhandled exception in spawn: Closed stream (IO::Error)
  from /usr/share/crystal/src/io.cr:128:5 in 'check_open'
  from /usr/share/crystal/src/io/buffered.cr:69:5 in 'read'
  from /usr/share/crystal/src/io.cr:530:20 in 'read_fully?'
  from /usr/share/crystal/src/io.cr:513:5 in 'read_fully'
  from /usr/share/crystal/src/io/byte_format.cr:131:3 in 'decode'
  from /usr/share/crystal/src/int.cr:520:5 in 'from_io'
  from /usr/share/crystal/src/io.cr:889:5 in 'read_bytes'
  from lib/pg/src/pq/connection.cr:115:7 in 'read_i32'
  from lib/pg/src/pq/connection.cr:177:7 in 'read_one_frame'
  from lib/pg/src/pq/connection.cr:169:31 in 'read_async_frame_loop'
  from lib/pg/src/pg/connection.cr:50:15 in '->'
  from /usr/share/crystal/src/fiber.cr:255:3 in 'run'
  from /usr/share/crystal/src/fiber.cr:48:34 in '->'
  from ???
```

`PQ::Connection#read_async_frame_loop` only swallows `Errno`, it should probably be extended to cover more errors, including `IO::Error`. Probably any error could be ignored if the connection is closed?

m-o-e:
+1

I'm now using this little monkey-patch to
be able to handle errors in `connect_listen`:

```crystal
class PG::Connection
  protected def listen
    @connection.read_async_frame_loop
  end
end
```

But needless to say, patching low-level
shards in this way doesn't feel good. ;)

Perhaps `connect_listen` could just be
a blocking call to begin with?

We can always wrap it in `spawn`
ourselves where needed and it gives much
better control over the error handling.

haarts:
I'm currently struggling a bit to actually make `connect_listen` blocking. I was somewhat surprised to learn it _didn't_ block. But perhaps that makes sense too.

carlhoerberg:
[...]

PR #238, opened 2021-11-10
Make listen blocking

carlhoerberg:
Exceptions, e.g. IO::EOFError, becasue the connection is lost are impossible to handle without this patch.

### DIFF
--- a/src/pg/connection.cr
+++ b/src/pg/connection.cr
@@ -47,7 +47,7 @@ module PG
     end
 
     protected def listen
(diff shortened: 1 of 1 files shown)


carlhoerberg:
FIxes https://github.com/will/crystal-pg/issues/195

will:
Thanks for submitting this, and reviving discussion on the issue, even if it wasn't merged. The other solution worked on in 195 has been released with v0.25.0

PR #241, opened 2021-12-18
Fix socket connections via url connection string

### LINKED ISSUE #227: Connecting to unix socket with url
I tried various forms such as
```
postgres://user:pass@/dbname
postgres://user:pass@/dbname?host=/var/run/postgresql
postgres://user:pass@%2Fvar%2Frun%2Fpostgresql/dbname
```
but they all doesn't seem to work

grepsedawk:
When connecting to a postgres instance via socket, the hostname parsed
by URI.parse is left empty, while a query string with the host is added
If the hostname is left blank, fallback to the host provided by the
query string

Resolves #227

### DIFF
--- a/spec/pq/conninfo_spec.cr
+++ b/spec/pq/conninfo_spec.cr
@@ -59,6 +59,13 @@ describe PQ::ConnInfo, ".from_conninfo_string" do
     assert_custom_params ci
   end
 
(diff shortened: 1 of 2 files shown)


robcole:
A+ branch name.

straight-shoota (approved):

will:
Thank you!

PR #242, opened 2021-12-30
Refactor `String#split` to `String#partition`

straight-shoota:
This is a performance improvement. `partition` doesn't allocate an array.

As a side effect, this also gets rid of rescuing `IndexError` (cf https://github.com/crystal-lang/crystal/issues/11565#issuecomment-1003040007).


will:
Nice

PR #243, opened 2022-01-14
Connect listen changes

will:
I think this addresses the things brought up in #195 by having connect_listen be optionally blocking and not add any breaking changes to this area.

### DIFF
--- a/src/pg.cr
+++ b/src/pg.cr
@@ -39,6 +39,7 @@ module PG
     # Close the connection.
     def close
       @conn.close
+    rescue
     end
   end
 end
--- a/src/pg.cr
+++ b/src/pg.cr
@@ -14,26 +14,35 @@ module PG
   #   pp notification.channel, notification.payload, notification.pid
   # end
   # ```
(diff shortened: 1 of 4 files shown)


m-o-e:
lgtm, thanks! :+1:

src/pg.cr
```diff
@@ -14,31 +14,41 @@ module PG
...
-    connect_listen(url, channels, &blk)
+  #
+  # By default, this will spawn a fiber to non-blocking listen. If you would
+  # rather handle this yourself, pass true to the blocking parameter.
+  #
+  # ```
+  # PG.connect_listen("postgres:///", "a", "b", blocking: true) { ... }
+  # ```
+
+  def self.connect_listen(url, *channels : String, blocking : Bool = false, &blk : PQ::Notification ->) : ListenConnection
```
bcardiff, line 25: LGTM, yet it is not clear to me that is users responsibility to close the underlying connection from the return value.

will: I think with the blocking case then it's on the user. Before it'd just stay open for the lifetime of the program and there was no real way to close it. But with this, it stays open until they close it.

PR #245, opened 2022-01-27
Update to support crystal-db 0.11.0

will:
ref: https://github.com/crystal-lang/crystal-db/pull/157

~I have the sha hard coded in just for now until there is a proper tag.~


src/pg/result_set.cr
```diff
@@ -76,6 +76,10 @@ class PG::ResultSet < ::DB::ResultSet
     decoder(index).type
   end
 
+  def next_column_index : Int32
+    @column_index
+  end
```
will, line 81: @bcardiff is this right? I'm not exactly sure the semantics for this new method, but it seems to pass the tests

bcardiff: Almost, a small tweak in the specs are needed, see 63431718f18d1f129cc03269b0a19a71f9a519bf. I can send the PR soon, or you can extract the changes from there. I was waiting on the db shared specs to be merged or not before sending the PR

will: Yeah just figured that out a few minutes ago :) I'm updating it now

bcardiff (approved):

PR #248, opened 2022-04-05
Update README.md

rdp:
Try to clarify why it mentions both crystal-db and also PG_DB methods.


rdp:
Note to self: after this I would like to add to readme how to do the correct requires... :)

will:
Thanks!

PR #254, opened 2022-10-04
Add support for UInt64 to decode LSN

louiseGrandjonc:


will:
@louiseGrandjonc can you rebase and repush? I switched from CircleCI to Github Actions since it looks like circleci broke

PR #256, opened 2022-10-12
Reuse existing connection for shared specs

will:
I noticed that the shared specs were making a new connection for every example, which wasn't necessary. Reusing the connection makes the tests to run on my laptop with compile time from an average of 12.275 seconds to 7.877

Nothing is failing now and I think the monkey patch is easy enough to back out should it start causing problems. But maybe there should be an official way to reuse an existing connection for the shared specs?


bcardiff (approved):
Sounds good. Adding a cleanup or "tables to be used" to DriverSpecs and a reuse: to with_db or something like that to crystal-db seems reasonable. 

I'll try to early next week most likely.

PR #262, opened 2022-10-22
Silence `_oid` argument warning

jgaskins:
This warning appears on the latest Crystal versions:

    In src/pg/decoder.cr:86:32

     86 | def decode(io, bytesize, _oid)
                                   ^---
    Warning: positional parameter '_oid' corresponds to parameter 'oid' of the overridden method PG::Decoders::Decoder#decode(io, bytesize, oid), which has a different name and may affect named argument passing

If the argument is expoosed outside the method as the name the Crystal compiler expects, this warning goes away. I kept the underscore-prefixed name as the internal variable name so it would remain clear it's not used.

### DIFF
--- a/src/pg/decoder.cr
+++ b/src/pg/decoder.cr
@@ -83,7 +83,7 @@ module PG
         2950, # UUID
       ]
 
-      def decode(io, bytesize, _oid)
+      def decode(io, bytesize, oid _oid)
         bytes = uninitialized UInt8[16]
 
         slice = Bytes.new(bytes.to_unsafe, 16)


jgaskins:
Looks like this was unneeded. I don't know what happened to my local repo. 🙃

will:
@jgaskins Thanks for opening this though, I had never seen giving arguments two names like that before though, that's neat :)

jgaskins:
It really is. I like the expressiveness it adds to method signatures like `def move(from source, to destination)` or `macro render(template, to io)`.

PR #264, opened 2023-03-09
Allow use of PG environment variables along with .pgpass files

wishdev:


will:
Thanks @wishdev! Could you though make one PR for adding the env vars and another for the pgpass? The reason being I think testing the env vars might be a little too much set up for CI, but the pgpass there could be some tests for. 

Maybe if the pgpass was split up a bit so the parsing could at least be tested, it'd help make sure it doesn't silently break in the future

wishdev:
@will I just pushed tests for all pieces - I did not break it up because I think testing everything covered your concerns here (if I read your comment correctly)

will:
Thanks!

PR #266, opened 2023-03-16
Decode TSTZ using connection's time zone

### LINKED ISSUE #265: Add ability to automatically return `TIMESTAMPTZ`->`Time` in the Postgres session's time zone
`TIMESTAMPTZ` columns are decoded without time zones, so I was trying to separate `TIMESTAMP` from `TIMESTAMPTZ` in [`TimeDecoder`](https://github.com/will/crystal-pg/blob/cafe7220c286fc5413b7aeebbec3d4c9efd06e62/src/pg/decoder.cr#L408-L434) but discovered that Postgres transmits them over the wire identically, without time zone info, and found [this comment](https://github.com/will/crystal-pg/pull/131#issuecomment-394092829) with a source that corroborates that.

The `psql` client automatically transforms timestamps into the correct time zone based on the session/transaction's `TIME ZONE` setting. My thought was that we could do the same here. Maybe after connecting we could run something like `@time_zone = query_one("SHOW TIME ZONE", as: String)`, and then `TimeDecoder` would run `time.in(time_zone)` for `TSTZ` columns, which would require passing the timezone somehow.

I'm not sure how we'd handle `SET LOCAL TIME ZONE` to set it for a transaction, though. We could maybe provide a `Transaction#time_zone=(location : Time::Location)` method, but I don't think the `crystal-db` shard supports driver-specific transactions yet so that would have to be a monkeypatch if we did it. If we don't provide a method for it, I don't know if we have a good way to detect the change.

jgaskins:
This is an idea for how #265 could be implemented, mostly implemented on my lunch break today. 🙃 

Closes #265

### DIFF
--- a/src/pg/connection.cr
+++ b/src/pg/connection.cr
@@ -2,6 +2,7 @@ require "../pq/*"
 
 module PG
   class Connection < ::DB::Connection
+    property time_zone = Time::Location.local
     protected getter connection
 
     def initialize(context)
--- a/src/pg/decoder.cr
+++ b/src/pg/decoder.cr
@@ -408,23 +408,29 @@ module PG
     struct TimeDecoder
       include Decoder
 
-      DATE_OID = 1082
+      enum OID
+        DATE        = 1082
+        TIMESTAMP   = 1114
+        TIMESTAMPTZ = 1184
+      end
       JAN_1_2K = Time.utc(2000, 1, 1)
 
-      def_oids [
-        DATE_OID, # date
-        1114,     # timestamp
-        1184,     # timestamptz
-      ]
+      def_oids OID.values.map(&.value)
 
       def decode(io, bytesize, oid)
-        if oid == DATE_OID
+        case oid = OID.new(oid)
+        in .date?
           v = read_i32(io)
           JAN_1_2K + Time::Span.new(days: v, hours: 0, minutes: 0, seconds: 0)
-        else
+        in .timestamp?
           v = read_i64(io) # microseconds
           sec, m = v.divmod(1_000_000)
           JAN_1_2K + Time::Span.new(seconds: sec, nanoseconds: m*1000)
+        in .timestamptz?
+          v = read_i64(io) # microseconds
+          sec, m = v.divmod(1_000_000)
+          time = JAN_1_2K + Time::Span.new(seconds: sec, nanoseconds: m*1000)
+          time.in io.connection.time_zone
         end
       end
 
--- a/src/pg/result_set.cr
+++ b/src/pg/result_set.cr
@@ -15,7 +15,7 @@ class PG::ResultSet < ::DB::ResultSet
     @column_index = -1 # The current column
     @end = false       # Did we read all the rows?
     @rows_affected = 0_i64
-    @sized_io = IO::Sized.new(conn.soc, 1)
+    @sized_io = Buffer.new(conn.soc, 1, statement.connection)
   end
 
   protected def conn
@@ -210,5 +210,13 @@ class PG::ResultSet < ::DB::ResultSet
   rescue DB::ConnectionLost
     # if the connection is lost there is nothing to be
     # done since the result set is no longer needed
+  end
+
+  private class Buffer < IO::Sized
+    getter connection : PG::Connection
+
+    def initialize(io, read_size, @connection)
+      super io, read_size
+    end
   end
 end
--- a/spec/pg/decoder_spec.cr
+++ b/spec/pg/decoder_spec.cr
@@ -53,6 +53,10 @@ describe PG::Decoders do
   test_decode "timestamptz", "'2015-02-03 16:15:14.23-01'::timestamptz",
     Time.utc(2015, 2, 3, 17, 15, 14, nanosecond: 230_000_000)
 
+  test_decode "timestamptz", "'2015-02-03 16:15:14.23 America/New_York'::timestamptz",
+    Time.local(2015, 2, 3, 16, 15, 14, nanosecond: 230_000_000, location: Time::Location.load("America/New_York")),
+    time_zone: Time::Location.load("America/New_York")
+
   test_decode "timestamp", "'2015-02-03 16:15:15'::timestamp",
     Time.utc(2015, 2, 3, 16, 15, 15)
 
--- a/spec/spec_helper.cr
+++ b/spec/spec_helper.cr
@@ -31,10 +31,23 @@ module Helper
   end
 end
 
-def test_decode(name, query, expected, file = __FILE__, line = __LINE__)
+def test_decode(name, query, expected, file = __FILE__, line = __LINE__, *, time_zone : Time::Location? = nil)
   it name, file, line do
-    value = PG_DB.query_one "select #{query}", &.read
-    value.should eq(expected), file: file, line: line
+    old_time_zone = Time::Location.local
+    PG_DB.using_connection do |c|
+      old_time_zone = c.time_zone
+      c = c.as PG::Connection
+      if time_zone
+        old_time_zone = c.time_zone
+        c.time_zone = time_zone
+      end
+      value = c.query_one "select #{query}", &.read
+      value.should eq(expected), file: file, line: line
+    ensure
+      if time_zone
+        c.time_zone = old_time_zone
+      end
+    end
   end
 end
 


src/pg/connection.cr
```diff
@@ -2,6 +2,7 @@ require "../pq/*"
 
 module PG
   class Connection < ::DB::Connection
+    property time_zone = Time::Location.local
```
jgaskins, line 5: I used local time here, but I figure for backwards compatibility we'd probably want it to be UTC?

src/pg/decoder.cr
```diff
@@ -408,23 +408,29 @@ module PG
...
+        TIMESTAMP   = 1114
+        TIMESTAMPTZ = 1184
+      end
       JAN_1_2K = Time.utc(2000, 1, 1)
 
-      def_oids [
-        DATE_OID, # date
-        1114,     # timestamp
-        1184,     # timestamptz
-      ]
```
jgaskins, line 418: This was just me cleaning up a bit, I figured OIDs were a fantastic use case for `enums.

src/pg/result_set.cr
```diff
@@ -211,4 +211,12 @@ class PG::ResultSet < ::DB::ResultSet
...
     # done since the result set is no longer needed
   end
+
+  private class Buffer < IO::Sized
+    getter connection : PG::Connection
+
+    def initialize(io, read_size, @connection)
+      super io, read_size
+    end
+  end
```
jgaskins, line 221: I don't know if this needed to be `private`, but this was the only backwards-compatible way I could come up with to pass the connection to the decoder.

src/pg/decoder.cr
```diff
@@ -408,23 +408,24 @@ module PG
...
-          sec, m = v.divmod(1_000_000)
-          JAN_1_2K + Time::Span.new(seconds: sec, nanoseconds: m*1000)
+        case oid = OID.new(oid)
+        in .date?
+          JAN_1_2K + read_i32(io).days
+        in .timestamp?
+          JAN_1_2K + read_i64(io).microseconds
+        in .timestamptz?
+          time = JAN_1_2K + read_i64(io).microseconds
+          time.in io.connection.time_zone
```
jgaskins, line 428: I also noticed we didn't need to `divmod` to split things up for the `Time::Span`. We could just use the microseconds value Postgres sends us and let the `Time::Span` constructor [take care of that work for us](https://github.com/crystal-lang/crystal/blob/29f9ac503fb028e1b095c8f8e036ed74b5474550/src/time/span.cr#L66-L84). It really simplifies this code — for my eyes, at least. 🙂

jgaskins:
One notable thing about this PR is that I didn't do anything with `SET TIME ZONE` or `SHOW TIME ZONE`. It's all being set in memory on the connection object. No idea if that's a good thing or a bad thing.

spec/pg/connection_spec.cr
```diff
@@ -104,3 +104,23 @@ describe PG, "#read_next_row_start" do
...
+      db.transaction do |txn|
+        db.exec "SET LOCAL TIME ZONE 'America/New_York'"
+        now = db.query_one "SELECT now()", as: Time
+        now.location.should eq Time::Location.load("America/New_York")
+      end
+
+      now = db.query_one "SELECT now()", as: Time
+      now.location.should eq Time::Location.load("America/Los_Angeles")
+    end
+  end
```
jgaskins, line 125: TIL server parameters (on the wire protocol, that is), so we don't have to store the time zone separately, and it will automatically be pushed to the client when `SET TIME ZONE` or `SET LOCAL TIME ZONE` is called. `PQ::Connection` already [stores all the server parameters as they come through](https://github.com/will/crystal-pg/blob/cafe7220c286fc5413b7aeebbec3d4c9efd06e62/src/pq/connection.cr#L223) so all we have to do is use it.

src/pg/connection.cr
```diff
@@ -41,6 +41,24 @@ module PG
...
+      if zone_name = @connection.server_parameters["TimeZone"]?
+        @@location_cache[zone_name]
+      else
+        Time::Location::UTC
+      end
+    end
+
+    def time_zone=(location : Time::Location)
+      exec "SET TIME ZONE '#{location.name}'"
+    end
```
jgaskins, line 60: I don't know if this method is necessary since I removed the `property` in https://github.com/will/crystal-pg/pull/266/commits/020d847bc5da15503f740e86809ec9455c6da165 to make it automatically handle `SET TIME ZONE` queries.

straight-shoota (commented):
When caching `Time::Location` it might make sense to consider a method for clearing the cache in case the time zone database is updated during the runtime of a program.

jgaskins:
I was trying to see what the performance penalty of decoding the `TSTZ` with the `Time::Location` is and it looks like it's about ~20µs of CPU time for 1k `TSTZ` values, or 20ns each. Here's the benchmark I used, where the only difference is whether it executes `time.in location` (the time calculation itself is [identical between `TS` and `TSTZ`](https://github.com/will/crystal-pg/pull/266/files#diff-f9ed9f723195ce49525495d296054748d6326d98e2e3a37c3de3f74f16d89331R424-R428)), fetching 1k time values from the DB on each query to amortize the latency cost:

```crystal
require "benchmark"
require "../src/pg"

db = DB.open("postgres:///")

Benchmark.ips do |x|
  x.report "timestamptz" { db.query_all "SELECT '2023-03-17 15:30 America/New_York'::timestamptz FROM generate_series(1, 1000)", as: Time }
  x.report "timestamp"   { db.query_all "SELECT '2023-03-17 15:30 America/New_York'::timestamp   FROM generate_series(1, 1000)", as: Time }
end

# This monkeypatch is taken from https://github.com/crystal-lang/crystal/pull/9846
# to measure CPU time used per iteration.
module Benchmark
  # Benchmark IPS calculates the number of iterations per second for a given
  # block of code. The strategy is to use two stages: a warmup stage and a
  # calculation stage.
  #
  # The warmup phase defaults to 2 seconds. During this stage we figure out how
  # many cycles are needed to run the block for roughly 100ms, and record it.
  #
  # The calculation defaults to 5 seconds. During this stage we run the block
  # in sets of the size calculated in the warmup stage. The measurements for
  # those sets are then used to calculate the mean and standard deviation,
  # which are then reported. Additionally we compare the means to that of the
  # fastest.
  module IPS
    class Job
      # List of all entries in the benchmark.
      # After #execute, these are populated with the resulting statistics.
      property items : Array(Entry)

      @warmup_time : Time::Span
      @calculation_time : Time::Span

      def initialize(calculation = 5, warmup = 2, interactive = STDOUT.tty?)
        @interactive = !!interactive
        @warmup_time = warmup.seconds
        @calculation_time = calculation.seconds
        @items = [] of Entry
      end

      # Adds code to be benchmarked
      def report(label = "", &action)
        item = Entry.new(label, action)
        @items << item
        item
      end

      def execute
        run_warmup
        run_calculation
        run_comparison
      end

      def report
        max_label = ran_items.max_of &.label.size
        max_compare = ran_items.max_of &.human_compare.size
        max_bytes_per_op = ran_items.max_of &.bytes_per_op.humanize(base: 1024).size

        ran_items.each do |item|
          printf "%s %s (%s real, %s user, %s sys) (±%5.2f%%)  %sB/op  %s\n",
            item.label.rjust(max_label),
            item.human_mean,
            item.human_iteration_time,
            item.human_cpu_user_time,
            item.human_cpu_system_time,
            item.relative_stddev,
            item.bytes_per_op.humanize(base: 1024).rjust(max_bytes_per_op),
            item.human_compare.rjust(max_compare)
        end
      end

      # The warmup stage gathers information about the items that is later used
      # in the calculation stage
      private def run_warmup
        @items.each do |item|
          GC.collect

          count = 0
          elapsed = Time.measure do
            target = Time.monotonic + @warmup_time

            while Time.monotonic < target
              item.call
              count += 1
            end
          end

          item.set_cycles(elapsed, count)
        end
      end

      private def run_calculation
        @items.each do |item|
          GC.collect

          measurements = [] of Time::Span
          bytes = 0_i64
          cycles = 0_i64
          cpu_user_seconds = 0.0_f64
          cpu_system_seconds = 0.0_f64

          target = Time.monotonic + @calculation_time

          loop do
            elapsed = nil
            before_cpu_time = nil
            after_cpu_time = nil

            bytes_taken = Benchmark.memory do
              before_cpu_time = Process.times
              elapsed = Time.measure { item.call_for_100ms }
              after_cpu_time = Process.times
              cpu_user_seconds += (after_cpu_time.utime + after_cpu_time.cutime) - (before_cpu_time.utime + before_cpu_time.cutime)
              cpu_system_seconds += (after_cpu_time.stime + after_cpu_time.cstime) - (before_cpu_time.stime + before_cpu_time.cstime)
            end
            bytes += bytes_taken
            cycles += item.cycles
            measurements << elapsed.not_nil!
            break if Time.monotonic >= target
          end

          ips = measurements.map { |m| item.cycles.to_f / m.total_seconds }
          item.calculate_stats(ips)

          item.bytes_per_op = (bytes.to_f / cycles.to_f).round.to_u64
          item.cpu_user_seconds = cpu_user_seconds / cycles
          item.cpu_system_seconds = cpu_system_seconds / cycles

          if @interactive
            run_comparison
            report
            print "\e[#{ran_items.size}A"
          end
        end
      end

      private def ran_items
        @items.select(&.ran?)
      end

      private def run_comparison
        fastest = ran_items.max_by { |i| i.mean }
        ran_items.each do |item|
          item.slower = (fastest.mean / item.mean).to_f
        end
      end
    end

    class Entry
      # Label of the benchmark
      property label : String

      # Code to be benchmarked
      property action : ->

      # Number of cycles needed to run for approx 100ms
      # Calculated during the warmup stage
      property! cycles : Int32

      # Number of 100ms runs during the calculation stage
      property! size : Int32

      # Statistical mean from calculation stage
      property! mean : Float64

      # Statistical variance from calculation stage
      property! variance : Float64

      # Statistical standard deviation from calculation stage
      property! stddev : Float64

      # Relative standard deviation as a percentage
      property! relative_stddev : Float64

      # Multiple slower than the fastest entry
      property! slower : Float64

      # Number of bytes allocated per operation
      property! bytes_per_op : UInt64

      # Number of seconds the CPU spent in executing userland code
      property! cpu_user_seconds : Float64

      # Number of seconds the CPU spent in system calls
      property! cpu_system_seconds : Float64

      @ran : Bool
      @ran = false

      def initialize(@label : String, @action : ->)
      end

      def ran?
        @ran
      end

      def call
        action.call
      end

      def call_for_100ms
        cycles.times { action.call }
      end

      def set_cycles(duration, iterations)
        @cycles = (iterations / duration.total_milliseconds * 100).to_i
        @cycles = 1 if cycles <= 0
      end

      def calculate_stats(samples)
        @ran = true
        @size = samples.size
        @mean = samples.sum.to_f / size.to_f
        @variance = (samples.reduce(0) { |acc, i| acc + ((i - mean) ** 2) }).to_f / size.to_f
        @stddev = Math.sqrt(variance)
        @relative_stddev = 100.0 * (stddev / mean)
      end

      def human_mean
        mean.humanize(precision: 2, significant: false, prefixes: Number::SI_PREFIXES_PADDED).rjust(7)
      end

      def human_iteration_time
        iteration_time = 1.0 / mean

        iteration_time.humanize(precision: 2, significant: false) do |magnitude, _|
          magnitude = Number.prefix_index(magnitude).clamp(-9..0)
          {magnitude, magnitude == 0 ? "s " : "#{Number.si_prefix(magnitude)}s"}
        end.rjust(8)
      end

      def human_cpu_user_time
        "#{cpu_user_seconds.humanize(precision: 2, significant: false)}s".rjust(8)
      end

      def human_cpu_system_time
        "#{cpu_system_seconds.humanize(precision: 2, significant: false)}s".rjust(8)
      end

      def human_compare
        if slower == 1.0
          "fastest"
        else
          sprintf "%5.2f× slower", slower
        end
      end
    end
  end
end
```

</details>

```
➜  crystal-pg git:(decode-tstz-with-timezone) crystal run --release examples/bench_time_zone_check.cr
timestamptz   5.81k (172.01µs real,  77.47µs user,   5.83µs sys) (± 0.81%)  58.7kB/op   1.11× slower
  timestamp   6.47k (154.62µs real,  55.48µs user,   7.01µs sys) (± 1.83%)  58.7kB/op        fastest
```

PR #269, opened 2023-05-31
bugfix: treat PGHOST env var as directory when it's a path

will:
This change makes the driver find the actual socket file and is consistent with how libpq appends .s.PGSQL.5432 to the value of the env var when it's a path:

    ❯ mkdir -p /tmp/foo && PGHOST=/tmp/foo psql
    psql: error: could not connect to server: No such file or directory
            Is the server running locally and accepting
            connections on Unix domain socket "/tmp/foo/.s.PGSQL.5432"?

When the value is a domain the behavior doesn't change

### DIFF
--- a/spec/pq/conninfo_spec.cr
+++ b/spec/pq/conninfo_spec.cr
@@ -132,5 +132,11 @@ describe PQ::ConnInfo, ".from_conninfo_string" do
       ci.port.should eq(1)
       ci.user.should eq("D")
     end
+
+    env_var_bubble do
+      ENV["PGHOST"] = "/path"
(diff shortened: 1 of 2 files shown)


ysbaddaden:
Documentation says that the socket is `.s.PGSQL.nnnn` where `nnnn` is the server port, so it's possible to start a Postgres server on the non default port, and that would affect the UNIX socket file? Maybe the PG shard should try to use `PGPORT` when specified?

will:
Yeah that looks like what `psql` using libpq does, so it's not unreasonable. Right now people can specify the path manually, but it could be nice to still get the automatic search to work too

PR #270, opened 2023-06-01
use nix for dev and ci

will:
This will make it easier to make sure CI runs the same thing as locally, and shouldn't have any bad effects for people who don't use nix.


shards.nix
```diff
@@ -0,0 +1,7 @@
+{
+  db = {
+    url = "https://github.com/crystal-lang/crystal-db.git";
+    rev = "v0.11.0";
```
bcardiff, line 4: What's the command for bumping this

will: If you are in the dev shell (either with `nix develop` or (what I do) have an .envrc file of `use flake`) then it's `crystal2nix` 

If not I think `nix develop -c crystal2nix` will just create a temp shell with crystal2nix in the path and then drop out of it right away

will:
Nice, the test caching is working as I had hoped: this commit (that will soon be squashed away) didn't touch any files that the specs depend on, so CI was basically just doing nothing
image
image

PR #274, opened 2023-10-03
Add support for application_name

cyberdelia:
Add support for `application_name` because not everything can be called crystal.


straight-shoota (approved):

will:
je te remercie

PR #276, opened 2023-11-05
Fix: statement.connection is a DB::Connection (expected PG::Connection)

ysbaddaden:
When requiring multiple database drivers (e.g. `mysql` or `sqlite3` in addition to `pg`), compilation will fail:

```
In lib/pg/src/pg/result_set.cr:18:51

 18 | @sized_io = Buffer.new(conn.soc, 1, statement.connection)
                                                    ^---------
Error: expected argument #3 to 'PG::ResultSet::Buffer.new' to be PG::Connection, not DB::Connection
```

As long as we don't require any other DB driver, `statement.connection` will always be `PG::Connection` as expected, but when requiring multiple drivers, it will be `DB::Connection+`. The solution is to make an explicit cast.

### DIFF
--- a/src/pg/result_set.cr
+++ b/src/pg/result_set.cr
@@ -15,7 +15,7 @@ class PG::ResultSet < ::DB::ResultSet
     @column_index = -1 # The current column
     @end = false       # Did we read all the rows?
     @rows_affected = 0_i64
-    @sized_io = Buffer.new(conn.soc, 1, statement.connection)
+    @sized_io = Buffer.new(conn.soc, 1, statement.connection.as(PG::Connection))
   end
 
   protected def conn


src/pg/result_set.cr
```diff
@@ -15,7 +15,7 @@ class PG::ResultSet < ::DB::ResultSet
     @column_index = -1 # The current column
     @end = false       # Did we read all the rows?
     @rows_affected = 0_i64
-    @sized_io = Buffer.new(conn.soc, 1, statement.connection)
+    @sized_io = Buffer.new(conn.soc, 1, statement.connection.as(PG::Connection))
```
bcardiff, line 18: There is a `conn` method that does this and is even used here in `conn.soc`. Doesn't this work?

```suggestion
    @sized_io = Buffer.new(conn.soc, 1, conn)
```

ysbaddaden: Almost, but nope:

```
Error: expected argument #3 to 'PG::ResultSet::Buffer.new' to be PG::Connection, not PQ::Connection
```

I tried to make Buffer take a PQ::Connection, but it breaks further down the line:

```
In lib/pg/src/pg/decoder.cr:428:33

 428 | time.in io.connection.time_zone
                             ^--------
Error: undefined method 'time_zone' for PQ::Connection
```

Last attempt, to no avail:

```
 18 | @sized_io = Buffer.new(conn.soc, 1, conn.as(PG::Connection))
                                          ^
Error: can't cast PQ::Connection to PG::Connection
```

bcardiff: Oh, PG vs PQ.

bcardiff (approved):

will:
Thanks!

PR #277, opened 2023-12-09
fix: timestamp(tz) arithmetic overflow

snadon:
A regression is causing an arithmetic overflow on some timestamp(tz)s.


bcardiff:
Wouldn't this be a bug in std-lib? Having a workaround here might make sense while that is updated, but the original seems it should work, right?

snadon:
Closing this issue since it will be resolved soon in std-lib.

will:
Thanks for finding the bug in the first place!

PR #282, opened 2024-01-22
fix: enable SNI

conradludgate:
👋 At Neon, we use SNI to know which database endpoint is being requested to connect to. Our fallback options are using `options` which this library also doesn't support, or using cleartext passwords and encoding the database endpoint into the password. See https://neon.tech/docs/connect/connection-errors#the-endpoint-id-is-not-specified

Reading https://forum.crystal-lang.org/t/specifying-the-sni-hostname-for-an-openssl-context/3872 it seems like enabling SNI wasn't too challenging. Testing locally it does seem to work.

### DIFF
--- a/src/pq/connection.cr
+++ b/src/pq/connection.cr
@@ -56,7 +56,7 @@ module PQ
         if sslrootcert = @conninfo.sslrootcert
           ctx.ca_certificates = sslrootcert
         end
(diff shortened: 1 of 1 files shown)


maattdd:
I confirm it works fine, could this be merged ? cc @bcardiff

bcardiff:
Merged! Thanks for the nudge

will:
Thanks @conradludgate! (and also thanks @bcardiff for getting to this before I did)

PR #283, opened 2024-03-08
Support encoding `PG::Interval` params

### LINKED ISSUE #272: Support inserting `PG::Interval` instances
I ran into an interesting case where in order to insert data in a prepared statement, you need to use `Time::Span`, but in order to read it out, as noted in the README, you need to use `PG::Interval`.

Is there a reason `PG::Interval` shouldn't delegate to `#to_span` to make it compatible in both directions? 

```sql
CREATE TABLE interval_testing
(
    id       BIGINT GENERATED ALWAYS AS IDENTITY PRIMARY KEY,
    duration INTERVAL NOT NULL
);
```
```cr
require "pg"

struct Testing
  include DB::Serializable

  getter id : Int64
  getter duration : PG::Interval
end

DB.open "postgres://administrator:4T5VohI8Zl6IySNemf@192.168.100.201:5432/reporting_warehouse" do |db|
  db.exec "INSERT INTO reporting.interval_testing (duration) VALUES ($1);", PG::Interval.new days: 1

  pp db.query_one "SELECT * FROM reporting.interval_testing WHERE id = 1;", as: Testing
end
```

Blacksmoke16:
Resolves #272.

I opted to encode the `PG::Interval` using the textual representation of a PG interval type, primarily to just have the most accuracy.

### DIFF
--- a/src/pq/param.cr
+++ b/src/pq/param.cr
@@ -68,6 +68,11 @@ module PQ
       text string
     end
 
+    def self.encode(val : PG::Interval)
+      # https://www.postgresql.org/docs/current/datatype-datetime.html#DATATYPE-INTERVAL-INPUT
+      text "#{val.months} months #{val.days} days #{val.microseconds} microseconds"
+    end
+
     def self.encode(val)
       text val.to_s
     end
--- a/spec/pg/encoder_spec.cr
+++ b/spec/pg/encoder_spec.cr
@@ -41,6 +41,9 @@ describe PG::Driver, "encoder" do
   test_insert_and_read "text[]", ["baz, bar"]
   test_insert_and_read "text[]", ["foo}"]
 
+  test_insert_and_read "interval", PG::Interval.new
+  test_insert_and_read "interval", PG::Interval.new(days: 400, microseconds: 5000000)
+
   describe "geo" do
     test_insert_and_read "point", PG::Geo::Point.new(1.2, 3.4)
     if Helper.db_version_gte(9, 4)


src/pq/param.cr
```diff
@@ -68,6 +68,11 @@ module PQ
       text string
     end
 
+    def self.encode(val : PG::Interval)
+      # https://www.postgresql.org/docs/current/datatype-datetime.html#DATATYPE-INTERVAL-INPUT
+      text "#{val.months} months #{val.days} days #{val.microseconds} microseconds"
```
Blacksmoke16, line 73: Alternatively, could move this logic into `PG::Interval#to_s(io)`?

jwoertink: How come no `minutes` or `seconds`? If I had something like this

```crystal
db.exec("INSERT INTO events (recurring_interval) VALUES (?)", args: 5.minutes)
```

Wouldn't this insert `0 months 0 days 0 microseconds`? Or is that unrelated to what this PR is for?

Blacksmoke16: I wondered about this too. But as far as I can tell the `PG::Interval` type only exposes the three properties I used. However your example would work just fine as `Time::Span` also works just fine for inserting.

I could see it making sense to also handle the opposite case of #272 where we allow reading `INTERVAL` PG types into a `Time::Span`. Then for most use cases you could use that, but if you really need the exact same behavior/precision you could use `PG::Interval`.

Blacksmoke16:
@will Any chance you could review this when you have some spare time?

PR #290, opened 2024-10-13
Handle `Array(Bytes)` encoding for query params containing non-Unicode text

### LINKED ISSUE #267: Exception sending query with bytea[] binary array type?
I ran into an `Unhandled exception: invalid byte sequence for encoding "UTF8": 0x00 (PQ::PQError)` when trying to do a query that involves the postgres `BYTEA` binary type. Any idea why it is trying to encode this as UTF8?

Simplified example code:

```crystal
array_of_binaries = Array(Bytes?).new
array_of_binaries << "Hello".to_slice
array_of_binaries << nil
array_of_binaries << "world".to_slice
array_of_binaries << Bytes[0, 255] # <=== This line breaks the query.

my_db.query("SELECT * FROM UNNEST($1::bytea[])", args: [array_of_binaries]) do |rs|
  rs.each do
    puts rs.read(Bytes?)
  end
end
```

As shown, I get this backtrace:

```
Unhandled exception: invalid byte sequence for encoding "UTF8": 0x00 (PQ::PQError)
  from lib/pg/src/pq/connection.cr:215:7 in 'handle_error'
  from lib/pg/src/pq/connection.cr:198:7 in 'handle_async_frames'
  from lib/pg/src/pq/connection.cr:174:7 in 'read'
  from lib/pg/src/pq/connection.cr:169:7 in 'read'
  from lib/pg/src/pq/connection.cr:447:31 in 'expect_frame'
  from lib/pg/src/pq/connection.cr:446:5 in 'expect_frame'
  from lib/pg/src/pg/statement.cr:19:5 in 'perform_query'
  from lib/db/src/db/statement.cr:93:9 in 'perform_query_with_rescue'
  from lib/db/src/db/statement.cr:80:7 in 'query:args'
  from lib/db/src/db/pool_statement.cr:29:30 in 'query:args'
  from lib/db/src/db/query_methods.cr:46:7 in 'query:args'
  from lib/db/src/db/query_methods.cr:61:7 in '__crystal_main'
```

If I comment out the line with the `Bytes[0, 255]`, then it works fine:

```
Bytes[72, 101, 108, 108, 111]

Bytes[119, 111, 114, 108, 100]
```

Any ideas on where I can look to fix this? [Postgres bytea docs](https://www.postgresql.org/docs/current/datatype-binary.html) suggest it can handle arbitrary binary data, so I suspect an issue on the crystal-pg side, but am not sure where to look. Thank you for any pointers.

jgaskins:
Ran into #267 tonight trying to bulk-insert MessagePack-encoded data (via `unnest`) and can confirm that it's `Array(Bytes)` that it's choking on. Since all arrays are text-encoded, we need to ensure text encoding for `Bytes` can handle binary data.

I tried special-casing `Array(Bytes)` all the way down, but not only does that require a lot more understanding of the binary encoding format than I currently have, it would've been an incomplete solution that only handles linear `bytea` arrays. Anyone using nested arrays would run into the same problem. So I decided against that.

I don't like that this solution increases the payload size over the wire by encoding every byte as 2 hexadecimal characters. Arrays of strings are implemented in terms of this method, so this impacts `text[]` encoding, but it doesn't impact encoding of strings _outside_ of arrays, so performance impact should be minimal for most use cases. And every other PQ implementation I've found so far, other than `libpq` itself, also uses the hex encoding so it seems that if we're slow we're at least in good company. I think my ideal solution would be to convert all param types to binary encoding, but this was a far easier step and I couldn't find the specification for the binary encoding and it's late and I'm sleeeeeeepy.

Please check my work on this. I did confirm that the test case deserializes into the expected string:

```crystal
db.query_each "SELECT '\\x7468697320697320612022736c69636522'::bytea" do |rs|
  puts String.new rs.read(Bytes)
end
# => this is a "slice"
```

… and my MessagePack-encoded data unpacks correctly after bulk-inserting it now, but I'm not confident there isn't something I've overlooked.

Fixes #267

### DIFF
--- a/spec/pq/param_spec.cr
+++ b/spec/pq/param_spec.cr
@@ -16,6 +16,7 @@ describe PQ::Param do
       it_encodes_array([1, 2, 3], "{1,2,3}")
       it_encodes_array([1.2, 3.4, 5.6], "{1.2,3.4,5.6}")
       it_encodes_array([%(a), %(\\b~), %(c\\"d), %(\uFF8F)], %({"a","\\\\b~","c\\\\\\"d","\uFF8F"}))
+      it_encodes_array([%(this is a "slice").to_slice], %({"\\\\x7468697320697320612022736c69636522"}))
       it_encodes_array(["baz, bar"], %({"baz, bar"}))
       it_encodes_array(["foo}"], %({"foo}"}))
       it_encodes_array([nil, nil], %({NULL,NULL}))
--- a/src/pq/param.cr
+++ b/src/pq/param.cr
@@ -121,8 +121,10 @@ module PQ
     end
 
     def self.encode_array(io, value : Bytes)
-      io << '"'
-      io << String.new(value).gsub(%("), %(\\"))
+      io << %{"\\\\x}
+      value.each do |byte|
+        byte.to_s io, base: 16, precision: 2
+      end
       io << '"'
     end
 


jgaskins:
> Arrays of strings are implemented in terms of this method, so this impacts `text[]` encoding

I just realized, this part may be a lie because [this spec](https://github.com/will/crystal-pg/pull/290/files#diff-1b70b376aa583c070e8738a9700d8035c47e8ddf2fb3ed79d8419e4589147b59R18) still passes. I must've read the code wrong.

will:
Thanks!!

Issue #291, opened 2024-10-13
 Error: undefined fun 'getuid' for LibC


jwoertink:
I'm working on getting [Avram Windows Support](https://github.com/luckyframework/avram/pull/1069) and when running on Windows, I got this error

![image](https://github.com/user-attachments/assets/faa67444-c159-4d1b-96e9-a8bdd5511180)

It looks like it's coming from here

https://github.com/will/crystal-pg/blob/cafeb43258bbfdde2bc7169e9d121b0bbe157dbe/src/pq/conninfo.cr#L174

This is running Crystal 1.14.0 on whatever `windows-latest` is.

Maybe related: https://github.com/crystal-lang/crystal/pull/12605

will:
Thanks for reporting this! The username is there to mimic the default behavior of libpq where by default it'll try to connect to a database with your current user's name and using a role named after your user. 

I figure there are two ways forward to fix this on Windows
1. just remove the default detection on windows and force database names and roles to be explicitly set.
2. Figure out how to get the current username on windows.

1 is a lot easier to do, especially without access to a windows computer and the tradeoff of being explicit doesn't feel _too_ onerous, but for sure 2 is a nicer experience over all. 

What do you think?

straight-shoota:
Here is the implementation in libpq: https://github.com/postgres/postgres/blob/7175ef870e24763a561821685160c5e35b2e8989/src/interfaces/libpq/fe-auth.c#L1186-L1245

will:
Thanks for finding that @straight-shoota . Also relevant https://github.com/postgres/postgres/blob/7175ef870e24763a561821685160c5e35b2e8989/src/interfaces/libpq/fe-auth.c#L1255-L1262 where for non-windows they get the user id to call that function you linked, but for windows it's ignored and so instead hardcoded to 0. 

I'm not sure what Windows header file `GetUserName` comes from. Searching comes up with a similar but not exactly the same named function https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-getusernamea . I do think `GetUserName` with no suffix is a windows api function given this 22 year old commit https://github.com/postgres/postgres/commit/469cb65aca0307506886158be2fce9f4438f8361 , but I'm afraid I really know next to nothing about Windows.

[...]

PR #292, opened 2024-10-31
windows: avoid breaking Libc.getuid call

will:
partially address #291

@jwoertink does this make it not crash at least? I don't have a windows install to test on myself

### DIFF
--- a/src/pq/conninfo.cr
+++ b/src/pq/conninfo.cr
@@ -171,7 +171,13 @@ module PQ
     end
 
     private def current_user_name
-      System::User.find_by(id: LibC.getuid.to_s).username
+      {% if flag?(:windows) %}
+        # NOTE: actually getting the current username on windows would be better
+        #       https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-getusernamew
+        "postgres"
+      {% else %}
+        System::User.find_by(id: LibC.getuid.to_s).username
+      {% end %}
     end
   end
 end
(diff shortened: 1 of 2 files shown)


jwoertink:
I think this might work... I'm now getting a different compiler error though...

I assume this isn't PG related.

```
Showing last frame. Use --error-trace for full trace.

In D:\a\_temp\crystal-latest-true-undefined\src\openssl\ssl\socket.cr:150:33

 150 | raise OpenSSL::SSL::Error.new(@ssl, bytes, "SSL_write")
                                 ^--
Error: Error: instance variable '@underlying_eof' of OpenSSL::SSL::Error must be Bool, not Nil
Error: Process completed with exit code 1.
```

src/pq/conninfo.cr
```diff
@@ -171,7 +171,13 @@ module PQ
     end
 
     private def current_user_name
-      System::User.find_by(id: LibC.getuid.to_s).username
+      {% if flag?(:windows) %}
+        # NOTE: actually getting the current username on windows would be better
+        #       https://learn.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-getusernamew
+        "crystal"
```
jwoertink, line 177: Instead of using `crystal` would `postgres` be a good alternative? Generally that username will exist. That's what we used as a fallback for Avram

https://github.com/luckyframework/avram/blob/b04ba2d4c130d6ddaf98069bfdedbf8a2b7b935a/src/avram.cr#L35

will: Yeah that makes sense!

jwoertink:
I've been able to confirm that the error I got is not related to PG. I've found a fix and confirmed this patch does fix the original issue letting Avram compile on Windows.

jwoertink (approved):

PR #295, opened 2024-12-01
Encode query params in binary instead of text

### LINKED ISSUE #294: Use binary encoding for bind arguments
Following up on #290, I would like to encode all bind params as binary. Partly because it might help avoid issues like #267 entirely to use a single encoding format, but also I ran a quick benchmark and, unless I've made a mistake in that benchmark, it's faster by a surprising margin and allocates significantly less on the heap (zero bytes allocated for `String` and `UUID` parameters).

<details><summary>benchmark code</summary>

```crystal
require "benchmark"
require "uuid"
require "../src/pq/param"
require "../src/pg/geo"

buffer = IO::Memory.new(1 << 16)

{
  1234,
  1234.5678,
  Time.utc,
  UUID.v7,
  "abc",           # short string
  "." * 20 * 1024, # 20KB
  PG::Geo::Point.new(-76.4218, 39.288436),
}.each do |value|
  puts value.class
  Benchmark.ips do |x|
    x.report "text" { buffer.rewind.write text_encode(value).slice }
    x.report "binary" { buffer.rewind.write binary_encode(value).slice }
  end
  puts
end

def text_encode(value : Int32 | Float64 | UUID | String | Time | PG::Geo::Point)
  PQ::Param.encode(value)
end

def binary_encode(value : Int32 | Int64 | Float64)
  slice = Bytes.new(sizeof(typeof(value)))
  IO::ByteFormat::NetworkEndian.encode value, slice
  PQ::Param.binary slice
end

def binary_encode(string : String)
  PQ::Param.binary string.to_slice
end

def binary_encode(uuid : UUID)
  PQ::Param.binary uuid.bytes.to_slice
end

def binary_encode(time : Time)
  binary_encode (time.to_unix_ns // 1_000).to_i64
end

def binary_encode(point : PG::Geo::Point)
  slice = Bytes.new(sizeof(Float64) * 2)
  IO::ByteFormat::NetworkEndian.encode point.x, slice
  IO::ByteFormat::NetworkEndian.encode point.y, slice + sizeof(Float64)
  PQ::Param.binary slice
end

```

</details>

<details><summary>benchmark results</summary>

```
➜  crystal-pg git:(master) ✗ crystal run --release examples/bench_encode.cr
Int32
  text  72.62M ( 13.77ns) (± 1.50%)  32.0B/op   1.54× slower
binary 111.73M (  8.95ns) (± 1.19%)  32.0B/op        fastest

Float64
  text  18.43M ( 54.27ns) (± 1.85%)   176B/op   6.12× slower
binary 112.74M (  8.87ns) (± 1.19%)  32.0B/op        fastest

Time
  text   7.87M (127.01ns) (± 1.54%)   176B/op  13.33× slower
binary 104.95M (  9.53ns) (± 1.21%)  16.0B/op        fastest

UUID
  text  27.56M ( 36.28ns) (± 0.83%)  176B/op   7.66× slower
binary 211.01M (  4.74ns) (± 1.07%)  0.0B/op        fastest

String
  text 214.13M (  4.67ns) (± 1.61%)  0.0B/op   1.07× slower
binary 228.46M (  4.38ns) (± 1.81%)  0.0B/op        fastest

String
  text   4.82M (207.49ns) (± 6.31%)  0.0B/op   1.01× slower
binary   4.87M (205.24ns) (± 7.41%)  0.0B/op        fastest

PG::Geo::Point
  text  11.30M ( 88.51ns) (± 1.74%)   144B/op   9.55× slower
binary 107.91M (  9.27ns) (± 0.68%)  32.0B/op        fastest
```

</details>

<details><summary>additional notes on relative performance</summary>

In fairness, this benchmark only measures the time spent encoding the data into a slice. Presumably, we spend a trivial amount of time encoding a handful of query params compared to the time spent in TLS, the kernel segments of the network stack, and decoding an unbounded quantity of result values.

The maximum number of bind parameters that Postgres supports in a single query is 65535, so I don't imagine this is likely to have a notable impact on performance. The only way to know for sure, though, is to benchmark holistically.

</details>

jgaskins:
Been messing with this the past couple days. I'm not sure yet, but this may actually not be a good idea. Encoding values as binary doesn't seem to get the benefit of autocasting, such as from `int4` up to `int8`. Either that or I've screwed something up.

I'll annotate some of the weirdness I saw in review comments. Maybe someone knows how to fix/work around those things.

It did cut down pretty significantly on the amount of heap memory allocated per query with bind args, though. It was a lot more than I expected for a simple query — about 44% fewer bytes allocated across the complete query execution. There are still some places I want to optimize heap allocations (like in serializing array types), but that was pretty nice to see.

```crystal
require "benchmark"
require "../src/pg"

pg = DB.open "postgres:///"
pg.exec <<-SQL
  CREATE TABLE IF NOT EXISTS temp_data (
    id UUID PRIMARY KEY,
    name TEXT NOT NULL,
    login_count INT4 NOT NULL,
    rating FLOAT8 NOT NULL,
    created_at TIMESTAMPTZ NOT NULL
  )
  SQL

Benchmark.ips do |x|
  time = Time.utc
  x.report do
    pg.exec <<-SQL, UUID.v7, "User Name", 42, rand, time
      INSERT INTO temp_data (id, name, login_count, rating, created_at)
      VALUES ($1, $2, $3, $4, $5)
      SQL
  end
end

at_exit { pg.exec "DROP TABLE IF EXISTS temp_data" }
```

</details>

On the `master` branch:

```
➜  crystal-pg git:(master) crystal run --release bench/encoding.cr
  30.88k ( 32.38µs) (± 1.64%)  992B/op  fastest
```

With this PR:

```
➜  crystal-pg git:(use-binary-encoding) crystal run --release bench/encoding.cr
  31.65k ( 31.59µs) (± 2.15%)  560B/op  fastest
```

Fixes #294

### DIFF
--- a/spec/pg/encoder_spec.cr
+++ b/spec/pg/encoder_spec.cr
@@ -6,7 +6,7 @@ enum EncoderSpec::Status
 end
 
 private def test_insert_and_read(datatype, value, file = __FILE__, line = __LINE__)
-  it "inserts #{datatype}", file, line do
+  it "inserts #{value.inspect} as #{datatype}", file, line do
     PG_DB.exec "drop table if exists test_table"
     PG_DB.exec "create table test_table (v #{datatype})"
     PG_DB.exec "insert into test_table values ($1)", args: [value]
@@ -18,11 +18,18 @@ private def test_insert_and_read(datatype, value, file = __FILE__, line = __LINE
 end
 
 describe PG::Driver, "encoder" do
-  test_insert_and_read "int4", 123
+  test_insert_and_read "boolean", true
+  test_insert_and_read "boolean", false
 
-  test_insert_and_read "float", 12.34
+  test_insert_and_read "int2", 123i16
+  test_insert_and_read "int4", 123i32
+  test_insert_and_read "int8", 123i64
+
+  test_insert_and_read "float4", 12.34f32
+  test_insert_and_read "float8", 12.34f64
 
   test_insert_and_read "varchar", "hello world"
+  test_insert_and_read "text", "hello world"
 
   test_insert_and_read "timestamp", Time.utc(2015, 2, 3, 17, 15, nanosecond: 13_000_000)
   test_insert_and_read "timestamp", Time.utc(2015, 2, 3, 17, 15, 13, nanosecond: 11_000_000)
@@ -32,18 +39,23 @@ describe PG::Driver, "encoder" do
   test_insert_and_read "int4", EncoderSpec::Status::Open
   test_insert_and_read "int4", EncoderSpec::Status::Closed
 
+  test_insert_and_read "bool[]", [] of Bool
+  test_insert_and_read "bool[]", [true]
+  test_insert_and_read "bool[]", [false]
   test_insert_and_read "bool[]", [true, false, true]
 
   test_insert_and_read "float[]", [1.2, 3.4, 5.6]
 
   test_insert_and_read "integer[]", [] of Int32
   test_insert_and_read "integer[]", [1, 2, 3]
-  test_insert_and_read "integer[]", [[1, 2], [3, 4]]
+  test_insert_and_read "integer[][]", [[1, 2], [3, 4]]
+  test_insert_and_read "integer[][][]", [[[1, 2], [3, 4]], [[5, 6], [7, 8]]]
 
   test_insert_and_read "text[]", ["t", "f", "t"]
   test_insert_and_read "text[]", [%("a), %(\\b~), %(c\\"d), %(\uFF8F)]
   test_insert_and_read "text[]", ["baz, bar"]
   test_insert_and_read "text[]", ["foo}"]
+  test_insert_and_read "text[][]", [["foo", "bar"], ["baz", "quux"]]
 
   test_insert_and_read "interval", PG::Interval.new
   test_insert_and_read "interval", PG::Interval.new(days: 400, microseconds: 5000000)
--- a/src/pg/decoder.cr
+++ b/src/pg/decoder.cr
@@ -501,7 +501,8 @@ module PG
       end
     end
 
-    @@decoders = Hash(Int32, PG::Decoders::Decoder).new(ByteaDecoder.new)
+    # :nodoc:
+    class_getter decoders = Hash(Int32, PG::Decoders::Decoder).new(ByteaDecoder.new)
 
     def self.from_oid(oid)
       @@decoders[oid]
--- a/src/pq/param.cr
+++ b/src/pq/param.cr
@@ -1,164 +1,292 @@
+require "../pg/interval"
 require "../pg/geo"
 
 module PQ
   # :nodoc:
-  record Param, slice : Slice(UInt8), size : Int32, format : Int16 do
+  record Param, slice : Slice(UInt8), size : Int32, format : Format do
+    enum Format : Int16
+      None   = -1
+      Text   =  0
+      Binary =  1
+    end
     delegate to_unsafe, to: slice
 
     #  Internal wrapper to represent an encoded parameter
 
     def self.encode(val : Nil)
-      binary Pointer(UInt8).null.to_slice(0), -1
+      binary Bytes.empty, -1
+    end
+
+    def self.encode(val : Bool, into slice : Bytes = Bytes.new(1))
+      slice[0] = val ? 1u8 : 0u8
+      binary slice
+    end
+
+    def self.encode(val : String)
+      encode val.to_slice
+    end
+
+    def self.encode(val : String, into slice : Bytes)
+      encode val.to_slice, into: slice
     end
 
     def self.encode(val : Slice)
-      binary val, val.size
+      binary val
     end
 
-    def self.encode(val : Array)
-      text encode_array(val)
+    def self.encode(val : Slice, into slice : Bytes)
+      val.copy_to slice
+
+      binary slice
+    end
+
+    def self.encode(val : Array(T)) forall T
+      bytes = ArrayEncoder.new(val).to_slice
+
+      binary bytes
     end
 
     def self.encode(val : Time)
-      text Time::Format::RFC_3339.format(val, fraction_digits: 9)
+      # text Time::Format::RFC_3339.format(val, fraction_digits: 9)
+      encode ((val - 30.years).to_unix_ns // 1_000).to_i64
     end
+
+    {% for type in %w[Int16 Int32 Int64 Float32 Float64] %}
+      def self.encode(val : {{type.id}}, into slice : Bytes = Bytes.new(sizeof(typeof(val))))
+        IO::ByteFormat::NetworkEndian.encode val, slice
+        # pp value: val, type: typeof(val), slice: slice
+        binary slice
+      end
+    {% end %}
 
     def self.encode(val : Enum)
       encode val.value
     end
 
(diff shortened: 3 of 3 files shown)


spec/pg/encoder_spec.cr
```diff
@@ -32,18 +39,23 @@ describe PG::Driver, "encoder" do
...
+  test_insert_and_read "bool[]", [false]
   test_insert_and_read "bool[]", [true, false, true]
 
   test_insert_and_read "float[]", [1.2, 3.4, 5.6]
 
   test_insert_and_read "integer[]", [] of Int32
   test_insert_and_read "integer[]", [1, 2, 3]
-  test_insert_and_read "integer[]", [[1, 2], [3, 4]]
+  test_insert_and_read "integer[][]", [[1, 2], [3, 4]]
+  test_insert_and_read "integer[][][]", [[[1, 2], [3, 4]], [[5, 6], [7, 8]]]
```
jgaskins, line 52: I was actually surprised that the `integer[]` example worked before with a nested array. It still does when encoded as binary, too, but I figured I would make this more explicit.

spec/pg/encoder_spec.cr
```diff
@@ -18,11 +18,18 @@ private def test_insert_and_read(datatype, value, file = __FILE__, line = __LINE
...
-  test_insert_and_read "float", 12.34
+  test_insert_and_read "int2", 123i16
+  test_insert_and_read "int4", 123i32
+  test_insert_and_read "int8", 123i64
+
+  test_insert_and_read "float4", 12.34f32
+  test_insert_and_read "float8", 12.34f64
 
   test_insert_and_read "varchar", "hello world"
+  test_insert_and_read "text", "hello world"
```
jgaskins, line 32: This was more of a sanity check than anything

spec/pg/encoder_spec.cr
```diff
@@ -18,11 +18,18 @@ private def test_insert_and_read(datatype, value, file = __FILE__, line = __LINE
 end
 
 describe PG::Driver, "encoder" do
-  test_insert_and_read "int4", 123
+  test_insert_and_read "boolean", true
+  test_insert_and_read "boolean", false
```
jgaskins, line 22: I noticed there weren't any tests for a plain boolean column, just an array of them. Probably didn't matter, but I needed to solve for booleans before solving for arrays of booleans. 😄

src/pg/decoder.cr
```diff
@@ -501,7 +501,8 @@ module PG
       end
     end
 
-    @@decoders = Hash(Int32, PG::Decoders::Decoder).new(ByteaDecoder.new)
+    # :nodoc:
+    class_getter decoders = Hash(Int32, PG::Decoders::Decoder).new(ByteaDecoder.new)
```
jgaskins, line 505: I did this because I was going to use decoders as the source for oids, but I didn't end up using it. I can probably undo this change.

src/pq/param.cr
```diff
@@ -1,164 +1,292 @@
...
     delegate to_unsafe, to: slice
 
     #  Internal wrapper to represent an encoded parameter
 
     def self.encode(val : Nil)
-      binary Pointer(UInt8).null.to_slice(0), -1
+      binary Bytes.empty, -1
+    end
+
+    def self.encode(val : Bool, into slice : Bytes = Bytes.new(1))
```
jgaskins, line 20: Being able to encode a value into an existing slice is important for arrays and also allowed me to encode things like intervals and geo types in terms of numeric types.

src/pq/param.cr
```diff
@@ -1,164 +1,292 @@
...
 
     def self.encode(val : Time)
-      text Time::Format::RFC_3339.format(val, fraction_digits: 9)
+      # text Time::Format::RFC_3339.format(val, fraction_digits: 9)
+      encode ((val - 30.years).to_unix_ns // 1_000).to_i64
     end
 
+    {% for type in %w[Int16 Int32 Int64 Float32 Float64] %}
+      def self.encode(val : {{type.id}}, into slice : Bytes = Bytes.new(sizeof(typeof(val))))
+        IO::ByteFormat::NetworkEndian.encode val, slice
```
jgaskins, line 56: Whoever added the ability encode numeric values directly into slices without an intermediate `IO` gets the world's biggest high-five from me.

src/pq/param.cr
```diff
@@ -1,164 +1,292 @@
...
+        # pp value: val, type: typeof(val), slice: slice
```
jgaskins, line 57:
```suggestion
```

src/pq/param.cr
```diff
@@ -1,164 +1,292 @@
+require "../pg/interval"
 require "../pg/geo"
 
 module PQ
   # :nodoc:
-  record Param, slice : Slice(UInt8), size : Int32, format : Int16 do
+  record Param, slice : Slice(UInt8), size : Int32, format : Format do
+    enum Format : Int16
+      None   = -1
```
jgaskins, line 8: This was because I misread the encoding for `nil`. I'll remove it.

```suggestion
```

spec/pg/encoder_spec.cr
```diff
@@ -32,18 +39,23 @@ describe PG::Driver, "encoder" do
   test_insert_and_read "int4", EncoderSpec::Status::Open
   test_insert_and_read "int4", EncoderSpec::Status::Closed
 
+  test_insert_and_read "bool[]", [] of Bool
+  test_insert_and_read "bool[]", [true]
+  test_insert_and_read "bool[]", [false]
   test_insert_and_read "bool[]", [true, false, true]
```
jgaskins, line 45: This was just some more sanity checks

src/pq/param.cr
```diff
@@ -1,164 +1,290 @@
...
+      PG::Geo::Polygon.name     => 604,  # polygon
+      PG::Geo::Box.name         => 603,  # box
+      PG::Geo::LineSegment.name => 601,  # lseg
+      PG::Geo::Line.name        => 628,  # line
+      PG::Geo::Circle.name      => 718,  # circle
+      JSON::Any.name            => 3802, # jsonb
+      Time.name                 => 1184, # timestamptz
+      Time::Span.name           => 1186, # interval
+      PG::Interval.name         => 1186, # interval
+    } of String => Int32
```
jgaskins, line 176: This is what I was hoping to avoid when making decoders accessible from the `PQ` namespace. I used type names as keys rather than the types themselves because I wanted to make it extensible, but I don't know how realistic that is.

I think that, in order to do this extensibly, it would require writing both encoders and decoders, because on top of the OID mapping, it also has to know the size of a given value to encode it, and I ended up making a `SIZE_MAP` below. I don't like that three different data structures have to know about the types available.

This is part of why I don't know whether this PR was a good idea. I may have bitten off more than I had the motivation to chew here and I'm not sure it's worth restructuring the shard to support binary encoding.

spec/pg/encoder_spec.cr
will: Good catch

will:
I like this in general. Not only for the less allocations on the crystal side, but I would have to imagine that postgres not having to parse as much text on the database side can be significant with write heavy workloads.

jgaskins:
That's a solid point. Scaling Postgres (especially for writes) is far more challenging than scaling out app servers when you hit a CPU threshold in Crystal app code.

PR #296, opened 2024-12-03
Add `PG::Numeric#to_big_d`

jgaskins:
Exploring using `decimal` columns for storing currency values and representing them with [`BigDecimal`](https://crystal-lang.org/api/1.14.0/BigDecimal.html) values in Crystal code, but then noticed `BigDecimal` support hasn't made it into this shard yet. When `PG::Numeric#to_big_r` was added, I don't think `BigDecimal` existed in the Crystal stdlib.

I copied most of the code from `to_big_r`, but rather than building up a power of 10 for the second argument based on the placement of the decimal point, we can just use the placement of the decimal point itself.

~A couple of the tests are failing and I'm not sure how to fix them yet. They look like trivial ones, but I spent too long trying to figure out how BigDecimal` works and I'm out of gas for the evening.~


jgaskins:
> A couple of the tests are failing and I'm not sure how to fix them yet.

j/k the expected outputs were wrong

bcardiff (approved):
Looks great.

Don't we also need a way to encode BigDecimal as a parameter?

will:
Thanks @jgaskins ! Could you also please add a changelog item?

jgaskins:
> Don't we also need a way to encode BigDecimal as a parameter?

For #295, absolutely. Until then, is [the catch-all](https://github.com/will/crystal-pg/blob/59d3dd91e6bd79384df1278d405862b027614e4e/src/pq/param.cr#L80-L82) sufficient?

> Could you also please add a changelog item?

Sure thing!

jgaskins:
Just checking in, is there anything else this PR needs before merging?

will:
Oh whoops sorry about letting this sit for so long! I forgot to check back in. Thanks for bumping it

PR #299, opened 2025-04-30
Add `pg` log source to log entries.

hugopl:
Even the only module emitting logs is named `PQ` because of postgres libpq, etc... IMO it makes more sense to use the shard name as log source, or maybe `pg.pq`, since most people consuming this log entry will thing that `pq` was a typo for `pg` 😅


straight-shoota (approved):

will:
Thanks!

PR #301, opened 2025-07-18
Subscribe to logical replication

jgaskins:
The basic idea is that you create a replication subscription with `PG.connect_replication`. It's similar to `PG.connect_listen` but, unlike a `LISTEN` connection where every message is the same type (it just yields strings), each message can be a different type. So instead of passing a block you pass a `PG::Replication::Handler` instance with `received` methods defined for receiving different types of replication messages:

```crystal
class MyHandler
  include PG::Replication::Handler

  # Override this to tell the Postgres server where you are in the WAL.
  # Make sure you `yield` in between any setup and teardown.
  def received(data : PG::Replication::XLogData, connection : PG::Replication::Connection)
    yield
    connection.last_wal_byte_applied = data.wal_end
    connection.last_wal_byte_flushed = data.wal_end
  end

  # New relation defined
  def received(relation : PG::Replication::Relation)
  end

  # Record updated
  def received(update : PG::Replication::Update)
  end
end

subscriber = PG.connect url,
  handler: MyHandler.new,
  publication_name: "my_publication",
  slot_name: "my_replication_slot"
```

The default is to do nothing when receiving a `WALMessage`.

Most of the work here happens in `XLogData` events, which contain the `WALMessage`s that people will actually care about. Everything else seems to be protocol meta. The objects passed to the handler's `received` methods are the `XLogData`'s `message` value and are decoded based on message formats specified in [the Postgres docs](https://www.postgresql.org/docs/current/protocol-logicalrep-message-formats.html).

There are still some things I need and/or want to do here:
- [x] Implement message types
  - [x] Begin
  - [x] Message
  - [x] Commit
  - [x] Origin
  - [x] Relation
  - [x] Type
  - [x] Insert
  - [x] Update
  - [x] Delete
  - [x] Truncate
- [ ] Allow decoding `TupleData` as other types (rather than raw `Bytes`)
- [ ] Allow starting the WAL from somewhere other than the beginning
- [x] Implement keepalive responses
- [x] Since logical replication requires a change to `postgresql.conf`, the specs need to be opt-in or enabled only when logical replication is enabled

The initial spec does successfully invoke the handler's methods for the `CREATE TABLE`, `INSERT`, and `UPDATE` commands.

### DIFF
--- a/src/pg/replication.cr
+++ b/src/pg/replication.cr
@@ -1,0 +1,549 @@
+module PG::Replication
+  module Handler
+    def received(frame)
+    end
+  end
+
+  class Connection
+    getter handler : Handler
+    getter publication_name : String
+    getter slot_name : String
+    @latest_wal = 0i64
+
+    def initialize(uri : URI | String, @handler, *, @publication_name, @slot_name, blocking : Bool = false)
+      if uri.is_a? String
+        uri = URI.parse(uri)
+      else
+        uri = uri.dup
+      end
+      query_params = uri.query_params
+      query_params["replication"] = "database"
+      uri.query_params = query_params
+      @conn = DB.connect(uri).as(PG::Connection)
+      @conn.listen_replication(
+        publication_name: publication_name,
+        slot_name: slot_name,
+        blocking: blocking,
+      ) do |frame|
+        received frame
+      end
+
+      spawn do
+        loop do
+          sleep 10.seconds
+          send_keepalive
+        end
+      end
+    end
+
+    def received(frame : CopyBoth)
+    end
+
+    def received(frame : CopyData)
+      received frame.data
+    end
+
+    def received(data : XLogData)
+      handler.received data.message
+    end
+
+    def received(keepalive : KeepAlive)
+      @latest_wal = keepalive.wal_end
+      send_keepalive if keepalive.response_expected?
+    end
+
+    # This shouldn't ever be received, but it can be represented in memory so we
+    # need to include it for completeness.
+    def received(response : KeepAliveResponse)
+      raise NotImplementedError.new("KeepAliveResponses are intended to be sent, not received")
+    end
+
+    def close
+      send_keepalive
+      @conn.close
+    end
+
+    def send_keepalive
+      write { send_keepalive! }
+    end
+
+    private def send_keepalive! : Nil
+      CopyData.new(
+        StandbyStatusUpdate.new(
+          last_wal_byte_received: @latest_wal,
+          last_wal_byte_flushed: @latest_wal,
+          last_wal_byte_applied: @latest_wal,
+        )
+      )
+        .to_io @conn.connection.soc
+
+      @conn.connection.soc.flush
+    end
+
+    private def write(&)
+      write_mutex.synchronize { yield }
+    end
+
+    getter write_mutex = Mutex.new
+  end
+
+  abstract struct Frame
+    def self.from_io(io : IO) : self
+      case byte = io.read_byte
+      when nil
+        raise IO::EOFError.new("Connection was unexpectedly terminated")
+      when 'W'
+        CopyBoth.new(io)
+      when 'd'
+        CopyData.new(io)
+      else
+        raise Error.new("Unexpected byte marker: 0x#{byte.to_s(16)} (#{byte.chr.inspect})")
+      end
+    end
+  end
+
+  struct CopyBoth < Frame
+    getter format : Format
+    getter column_formats : Array(Int16)
+
+    def initialize(io : IO)
+      size = io.read_bytes(Int32, IO::ByteFormat::NetworkEndian)
+      sized = IO::Sized.new(io, size - 4)
+      @format = Format.new(sized.read_bytes(Int8, IO::ByteFormat::NetworkEndian))
+      column_count = sized.read_bytes(Int16, IO::ByteFormat::NetworkEndian)
+      @column_formats = Array.new(column_count) do
+        sized.read_bytes(Int16, IO::ByteFormat::NetworkEndian)
+      end
+      sized.close
+    end
+
+    enum Format : Int8
+      Text   = 0
+      Binary = 1
+    end
+  end
+
+  struct CopyData < Frame
+    getter data : XLogData | KeepAlive | KeepAliveResponse
+
+    def initialize(io : IO)
+      size = io.read_bytes(Int32, IO::ByteFormat::NetworkEndian)
+      case byte = io.read_byte
+      when Nil
+        raise IO::EOFError.new("Connection was unexpectedly terminated")
+      when 'w'
+        @data = XLogData.new(io)
+      when 'k'
+        @data = KeepAlive.new(io)
+      else
+        raise Error.new("Unexpected CopyData byte marker: 0x#{byte.to_s(16)} (#{byte.chr.inspect})")
+      end
+    end
+
+    def initialize(@data)
+    end
+
+    def to_io(io : IO) : Nil
+      buffer = IO::Memory.new
+      io << 'd'
+      payload = IO::Memory.new.tap { |buf| data.to_io buf }.to_slice
+      io.write_bytes payload.bytesize + 4, IO::ByteFormat::NetworkEndian
+      io.write payload
+    end
+  end
+
+  struct XLogData
+    getter wal_start : Int64
+    getter wal_end : Int64
+    getter timestamp : Time
+    getter message : WALMessage
+
+    def initialize(io : IO)
+      @wal_start = io.read_bytes(Int64, IO::ByteFormat::NetworkEndian)
+      @wal_end = io.read_bytes(Int64, IO::ByteFormat::NetworkEndian)
+      @timestamp = TimeParser.call(io)
+      @message = WALMessage.new(io)
+    end
+
+    def to_io(io : IO) : Nil
+      raise NotImplementedError.new("XLogData messages are meant to be sent by the server, not received by the client")
+    end
+  end
+
+  struct KeepAlive
+    getter wal_end : Int64
+    getter timestamp : Time
+    getter? response_expected : Bool
+
+    def initialize(io : IO)
+      @wal_end = io.read_bytes(Int64, IO::ByteFormat::NetworkEndian)
+      @timestamp = TimeParser.call(io)
+      @response_expected = io.read_bytes(Int8, IO::ByteFormat::NetworkEndian) == 1
+    end
+
+    def to_io(io : IO) : Nil
+      raise NotImplementedError.new("KeepAlives are intended to be received from the server, not sent to it")
+    end
+  end
+
+  abstract struct KeepAliveResponse
+  end
+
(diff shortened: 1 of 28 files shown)


src/pg/replication.cr
```diff
@@ -0,0 +1,353 @@
+module PG::Replication
```
jgaskins, line 6: I can't tell yet if this should be `LogicalReplication` or if physical replication uses the same concepts.

src/pg/replication.cr
```diff
@@ -0,0 +1,353 @@
...
+  module Handler
+    def received(frame)
+    end
```
jgaskins, line 51: This is needed so that you don't get compile-time errors when you don't define all possible messages. It doesn't have to be untyped, though. It might be better to be explicit that this is for `WALMessage` objects only.

src/pg/replication.cr
```diff
@@ -0,0 +1,353 @@
...
+        publication_name: publication_name,
+        slot_name: slot_name,
+        blocking: blocking,
+      ) do |frame|
+        received frame
+      end
+    end
+
+    def received(frame : CopyBoth)
+    end
```
jgaskins, line 175: I don't know what to do here or if there's anything _to_ do. This is the first message that comes in after you send the `START_REPLICATION` command.

src/pg/replication.cr
```diff
@@ -0,0 +1,353 @@
...
+
+    def received(frame : CopyData)
+      received frame.data
+    end
```
jgaskins, line 180: This is the message that contains the `XLogData`

src/pg/replication.cr
```diff
@@ -0,0 +1,353 @@
...
+
+    def received(data : XLogData)
+      handler.received data.message
+    end
```
jgaskins: … and the `XLogData` contains the `WALMessage`.

src/pg/replication.cr
```diff
@@ -0,0 +1,353 @@
...
+        raise Error.new("Unexpected WAL message byte marker: 0x#{byte.to_s(16)} (#{byte.chr.inspect})")
+      end
+    end
+
+    protected def read(io : IO, int : Int.class)
+      io.read_bytes int, IO::ByteFormat::NetworkEndian
+    end
+
+    protected def read_string(io : IO) : String
+      io.read_line('\0', chomp: true)
```
jgaskins: This usage makes me think `read_line` is a misnomer for this method. It's reading up to a delimiter.

This may also be the least efficient part of this whole PR. Reading a string up to a delimiter may involve internal reallocations for longer strings as the buffer keeps filling up.

src/pg/replication.cr
```diff
@@ -0,0 +1,353 @@
...
+    protected def read_tuple_data(io) : TupleData
+      column_count = read(io, Int16)
+      Array.new(column_count) do
+        case byte = io.read_byte
+        when Nil
+          raise IO::EOFError.new("Connection was unexpectedly terminated")
+        when 'n'
+          nil
+        when 'u'
+          UnchangedTOASTValue.new
```
jgaskins: I haven't tested this yet, but the docs say that [an unchanged TOASTed value is not sent](https://www.postgresql.org/docs/current/protocol-logicalrep-message-formats.html#PROTOCOL-LOGICALREP-MESSAGE-FORMATS-TUPLEDATA), so I'm assuming it's just a marker to indicate that the previous value is unchanged to avoid sending huge strings over the wire unnecessarily.

src/pq/connection.cr
```diff
@@ -188,45 +189,44 @@ module PQ
       end
     end
 
+    def start_replication_frame_loop(publication_name : String, slot_name : String, &block : PG::Replication::Frame ->)
+      command = "START_REPLICATION SLOT #{slot_name} LOGICAL 0/0 (proto_version '1', binary 'true', publication_names '#{publication_name}')"
```
jgaskins: A few notes on this line:

- There are currently [4 `proto_version` values](https://www.postgresql.org/docs/current/protocol-logical-replication.html#PROTOCOL-LOGICAL-REPLICATION-PARAMS), but anything other than `1` requires Postgres 14+ and changes how some `WALMessage`s are decoded. Supporting these dynamically would be more complicated, so I'm just supporting the baseline out of the gate.
- This is currently hardcoded to start at the beginning of the WAL (`0/0`). The `PG.listen_replication` method should almost certainly be updated to allow passing in other LSNs to start from a different point in the WAL history.

src/pq/connection.cr
```diff
@@ -188,45 +189,44 @@ module PQ
...
 
-    private def handle_notification(frame : Frame::NotificationResponse)
-      notification_handler.call(frame.as_notification)
+    private def handle_async_frames(frame : Frame::NoticeResponse)
+      notice_handler.call(frame.as_notice)
+      true
     end
 
-    private def handle_parameter(frame : Frame::ParameterStatus)
+    private def handle_async_frames(frame : Frame::ParameterStatus)
```
jgaskins, line 236: The reason I updated this was because I was originally going to try to reuse `read_async_frame_loop`. That didn't work, but along the way I realized that the `if frame.is_a?` was basically a `case` statement, and even that was equivalent to just using an overloaded method. When I backed out my attempt, I just left it because I think this refactor still simplifies the code a bit.

src/pq/connection.cr
```diff
@@ -239,18 +239,21 @@ module PQ
...
+    def connect(*, replication : String? = nil)
       startup_args = [
         "user", @conninfo.user,
         "database", @conninfo.database,
         "application_name", @conninfo.application_name,
         "client_encoding", "utf8",
       ]
+      if replication
+        startup_args << "replication" << replication
+      end
```
jgaskins, line 263: `["replication", "true"]` enables physical-replication mode and `["replication", "database"]` enables logical-replication mode ([docs](https://www.postgresql.org/docs/current/protocol-replication.html)). `PG::Replication::Connection` [currently hardcodes it to `"database"`](https://github.com/will/crystal-pg/pull/301/files#diff-9db8c24b86773bc96353706cc4aec1b44c47b8ee0e06c8a2828d818b9605e004R19) to enable logical replication.

jgaskins (commented):
Just some notes about the current state of things. Some of them are subject to change.

src/pq/connection.cr
will: FWIW postgres 13 is EOL in 3 months I think. Still starting support with only 1 is probably fine.

will:
I've not done much with logical replication myself, so I cant comment on the details of the protocol here, but it all seems super cool!

src/pq/connection.cr
jgaskins: Nice. Main addition of in v2 is transaction ids on various actions, which could be useful for some that might want this functionality. I wouldn't say it's necessary since the events are linear (all CRUD actions map to the transaction id from the most recent `Begin` WAL message), but it might still be useful.

jgaskins:
I've had to add Debezium to our stack to pull some data into our data warehouse. All the trouble I've had getting it working got me wondering how easy or difficult it would be to just consume the WAL ourselves and emit a similar output structure. _Completely_ nerd-sniped.

> I cant comment on the details of the protocol here

I'm hoping nobody will need to be all that concerned with it. 😄 It's not *super* complicated, but I want to come up with a good set of tests that illustrate that replication is working/broken by just performing actions and asserting on things appearing in the stream of logical replication events.

jgaskins:
Alrighty, I think this is at least ready for initial review. Feel free to let me know if anything is unclear.

will:
Awesome, thanks! I think this is all good to go. Could you add a changelog entry too please first?

jgaskins:
Sweet, will do. I did find a bug the other night, though, that I still need to address. Specifically: error frames aren't being decoded on this connection. While looking into that, I realized the error frame is [this error](https://www.postgresql.org/docs/current/protocol-message-formats.html#PROTOCOL-MESSAGE-FORMATS-ERRORRESPONSE), which you've already got here. Same message structure, which makes sense.

Then I wondered, since that's the same as `ErrorResponse`, if the [`CopyBoth` message](https://github.com/will/crystal-pg/pull/301/files#diff-9db8c24b86773bc96353706cc4aec1b44c47b8ee0e06c8a2828d818b9605e004R158-R177) is documented on that page, too. And [it is](https://www.postgresql.org/docs/current/protocol-message-formats.html#PROTOCOL-MESSAGE-FORMATS-COPYBOTHRESPONSE). That's when I realized I _could've_ reused the same connection structure as `LISTEN` like I originally intended.

I think, for the moment, I'll just reimplement the error frame here (the pattern I used in implementing these messages turned out pretty decent) and refactor toward reuse later.

jgaskins:
I've been running this PR in production to replace Debezium with a custom CDC implementation and it's been going pretty well. Last night it did start running into disconnection issues under sustained heavy write load, but that was fixed in these commits:

- https://github.com/will/crystal-pg/pull/301/commits/aad0585ac40d0245fa10099a349abd1c23e91541 adds automatic reconnect until explicitly closed
  - I'm not sure it's is the *correct* solution but, all things considered, it seems like a reasonable tradeoff. The `PG::Replication::Connection` instance tracks WAL cursors that haven't been sent to the server yet, so ideally we recover precisely where we left off.
- https://github.com/will/crystal-pg/pull/301/commits/de7771ffc65bb2ad39777903ebd98a8127abc887 makes the Postgres keepalive interval configurable
  - CloudNativePG sets `wal_sender_timeout` to [5 seconds](https://cloudnative-pg.io/docs/1.25/release_notes/edb-cloud-native-postgresql#:~:text=During%20a%20failover%2C%20wait%20to%20select%20a%20new%20primary%20until%20all%20the%20WAL%20streaming%20connections%20are%20closed.%20The%20operator%20now%20sets%20by%20default%20wal_sender_timeout%20and%20wal_receiver_timeout%20to%205%20seconds%20to%20make%20sure%20standby%20nodes%20will%20quickly%20notice%20if%20the%20primary%20has%20network%20issues) (from the Postgres default of 60 seconds) in order to reduce failover latency. That's less than the 10 seconds I'd hardcoded and this is why you don't use magic numbers.
  - A CPU-bound WAL handler won't guarantee the interval because Crystal's fiber scheduler doesn't preempt, so the WAL consumer's handler code needs to yield if it's doing a lot of work.

Even though the second commit fixed the disconnection issue I was seeing, getting disconnected for _any_ reason shouldn't require us to re-run WAL data we've already processed, so the first commit is still helpful.

After deploying these two changes, I ran the heavy write load again and we're getting CPU usage of 0.7-1.3% and RAM usage of 4MB on my Crystal CDC pod. I thought surely something was broken because that's far fewer resources than I expected under load (and certainly less than Debezium was consuming) but the data's appearing downstream so it's just surprisingly efficient.

will:
Oh wow that sounds great! Let's get this in then. I plan to press the squash and merge button, but let me know if you'd rather move things around yourself?

jgaskins:
I went ahead and touched up the commit history to boil the original 28 or so original commits down to 5 meaningful changes.

will:
Thanks!!

PR #304, opened 2025-11-14
add support for PostgreSQL range and multirange types

mamantoha:
This PR adds support for PostgreSQL range and multirange types.

### Features Added

- Supports all 6 built-in range types:
  `int4range`, `int8range`, `daterange`, `tsrange`, `tstzrange`, `numrange`
- Supports all 6 built-in multirange types on PostgreSQL 14+:
  `int4multirange`, `int8multirange`, `datemultirange`, `tsmultirange`, `tstzmultirange`, `nummultirange`
- Decodes ranges to `PG::Range(T)` to preserve PostgreSQL boundary semantics:
  inclusive/exclusive bounds, empty ranges, and infinite bounds
- Decodes multiranges to `Array(PG::Range(T))`
- Supports encoding `PG::Range(T)` values as query parameters
- Keeps native Crystal `Range` parameter encoding as a convenience for callers that do not need exclusive lower bounds

### Notes

PostgreSQL continuous range types such as `tsrange`, `tstzrange`, and `numrange` can distinguish `(a,b)` from `[a,b)`. Native Crystal `Range` cannot represent exclusive lower bounds, so this PR introduces `PG::Range(T)` for decoded values and exact parameter encoding.

Discrete range types such as `int4range`, `int8range`, and `daterange` are still canonicalized by PostgreSQL, for example `[1,10]` becomes `[1,11)`.

### Testing

- Added decoder coverage for all supported range types and boundary combinations
- Added encoder coverage for range and multirange query parameters
- Added regression coverage for preserving exclusive lower bounds on continuous ranges
- Covered empty ranges, infinite bounds, timezone handling, PostgreSQL canonicalization, and round-trip behavior

### Breaking Changes

None. This is a pure addition.

### DIFF
--- a/spec/pg/decoders/range_decoder_spec.cr
+++ b/spec/pg/decoders/range_decoder_spec.cr
@@ -36,8 +36,8 @@ describe PG::Decoders do
     end
 
     describe "tsrange" do
-      test_decode "(lower,upper) - both exclusive", "'(2023-01-01 10:30:00,2023-12-31 15:45:00)'::tsrange", Time.utc(2023, 1, 1, 10, 30, 0)...Time.utc(2023, 12, 31, 15, 45, 0)
-      test_decode "(lower,upper] - exclusive lower, inclusive upper", "'(2023-01-01 10:30:00,2023-12-31 15:45:00]'::tsrange", Time.utc(2023, 1, 1, 10, 30, 0)..Time.utc(2023, 12, 31, 15, 45, 0) # Crystal can't represent exclusive lower
+      test_decode "(lower,upper) - both exclusive", "'(2023-01-01 10:30:00,2023-12-31 15:45:00)'::tsrange", PG::Range.new(Time.utc(2023, 1, 1, 10, 30, 0), Time.utc(2023, 12, 31, 15, 45, 0), lower_inclusive: false)
+      test_decode "(lower,upper] - exclusive lower, inclusive upper", "'(2023-01-01 10:30:00,2023-12-31 15:45:00]'::tsrange", PG::Range.new(Time.utc(2023, 1, 1, 10, 30, 0), Time.utc(2023, 12, 31, 15, 45, 0), lower_inclusive: false, upper_inclusive: true)
       test_decode "[lower,upper) - inclusive lower, exclusive upper", "'[2023-01-01 10:30:00,2023-12-31 15:45:00)'::tsrange", Time.utc(2023, 1, 1, 10, 30, 0)...Time.utc(2023, 12, 31, 15, 45, 0)
       test_decode "[lower,upper] - both inclusive", "'[2023-01-01 10:30:00,2023-12-31 15:45:00]'::tsrange", Time.utc(2023, 1, 1, 10, 30, 0)..Time.utc(2023, 12, 31, 15, 45, 0)
       test_decode "empty range", "'empty'::tsrange", Time.unix(0)...Time.unix(0)
@@ -59,8 +59,8 @@ describe PG::Decoders do
     end
 
     describe "tstzrange" do
-      test_decode "(lower,upper) - both exclusive", "'(2023-01-01 10:30:00+00,2023-12-31 15:45:00+00)'::tstzrange", Time.utc(2023, 1, 1, 10, 30, 0)...Time.utc(2023, 12, 31, 15, 45, 0)
-      test_decode "(lower,upper] - exclusive lower, inclusive upper", "'(2023-01-01 10:30:00+00,2023-12-31 15:45:00+00]'::tstzrange", Time.utc(2023, 1, 1, 10, 30, 0)..Time.utc(2023, 12, 31, 15, 45, 0)
+      test_decode "(lower,upper) - both exclusive", "'(2023-01-01 10:30:00+00,2023-12-31 15:45:00+00)'::tstzrange", PG::Range.new(Time.utc(2023, 1, 1, 10, 30, 0), Time.utc(2023, 12, 31, 15, 45, 0), lower_inclusive: false)
+      test_decode "(lower,upper] - exclusive lower, inclusive upper", "'(2023-01-01 10:30:00+00,2023-12-31 15:45:00+00]'::tstzrange", PG::Range.new(Time.utc(2023, 1, 1, 10, 30, 0), Time.utc(2023, 12, 31, 15, 45, 0), lower_inclusive: false, upper_inclusive: true)
       test_decode "[lower,upper) - inclusive lower, exclusive upper", "'[2023-01-01 10:30:00+00,2023-12-31 15:45:00+00)'::tstzrange", Time.utc(2023, 1, 1, 10, 30, 0)...Time.utc(2023, 12, 31, 15, 45, 0)
       test_decode "[lower,upper] - both inclusive", "'[2023-01-01 10:30:00+00,2023-12-31 15:45:00+00]'::tstzrange", Time.utc(2023, 1, 1, 10, 30, 0)..Time.utc(2023, 12, 31, 15, 45, 0)
       test_decode "empty range", "'empty'::tstzrange", Time.unix(0)...Time.unix(0)
@@ -70,8 +70,8 @@ describe PG::Decoders do
     end
 
     describe "numrange" do
-      test_decode "(lower,upper) - both exclusive", "'(1.5,10.75)'::numrange", PG::Numeric.new(2_i16, 0_i16, 0_i16, 1_i16, [1_i16, 5000_i16])...PG::Numeric.new(2_i16, 0_i16, 0_i16, 2_i16, [10_i16, 7500_i16])
-      test_decode "(lower,upper] - exclusive lower, inclusive upper", "'(1.5,10.75]'::numrange", PG::Numeric.new(2_i16, 0_i16, 0_i16, 1_i16, [1_i16, 5000_i16])..PG::Numeric.new(2_i16, 0_i16, 0_i16, 2_i16, [10_i16, 7500_i16])
+      test_decode "(lower,upper) - both exclusive", "'(1.5,10.75)'::numrange", PG::Range.new(PG::Numeric.new(2_i16, 0_i16, 0_i16, 1_i16, [1_i16, 5000_i16]), PG::Numeric.new(2_i16, 0_i16, 0_i16, 2_i16, [10_i16, 7500_i16]), lower_inclusive: false)
+      test_decode "(lower,upper] - exclusive lower, inclusive upper", "'(1.5,10.75]'::numrange", PG::Range.new(PG::Numeric.new(2_i16, 0_i16, 0_i16, 1_i16, [1_i16, 5000_i16]), PG::Numeric.new(2_i16, 0_i16, 0_i16, 2_i16, [10_i16, 7500_i16]), lower_inclusive: false, upper_inclusive: true)
       test_decode "[lower,upper) - inclusive lower, exclusive upper", "'[1.5,10.75)'::numrange", PG::Numeric.new(2_i16, 0_i16, 0_i16, 1_i16, [1_i16, 5000_i16])...PG::Numeric.new(2_i16, 0_i16, 0_i16, 2_i16, [10_i16, 7500_i16])
       test_decode "[lower,upper] - both inclusive", "'[1.5,10.75]'::numrange", PG::Numeric.new(2_i16, 0_i16, 0_i16, 1_i16, [1_i16, 5000_i16])..PG::Numeric.new(2_i16, 0_i16, 0_i16, 2_i16, [10_i16, 7500_i16])
       test_decode "empty range", "'empty'::numrange", PG::Numeric.new(1_i16, 0_i16, 0_i16, 0_i16, [0_i16])...PG::Numeric.new(1_i16, 0_i16, 0_i16, 0_i16, [0_i16])
--- a/spec/pg/decoders/range_decoder_spec.cr
+++ b/spec/pg/decoders/range_decoder_spec.cr
@@ -1,0 +1,109 @@
+require "../../spec_helper"
+
+describe PG::Decoders do
+  describe "ranges" do
+    describe "int4range" do
+      test_decode "(lower,upper) - both exclusive", "'(1,10)'::int4range", 2...10 # PostgreSQL canonicalizes discrete ranges to [a,b) form
+      test_decode "(lower,upper] - exclusive lower, inclusive upper", "'(1,10]'::int4range", 2...11
+      test_decode "[lower,upper) - inclusive lower, exclusive upper", "'[1,10)'::int4range", 1...10
+      test_decode "[lower,upper] - both inclusive", "'[1,10]'::int4range", 1...11 # [a,b] becomes [a,b+1) for discrete types
+      test_decode "empty range", "'empty'::int4range", 0...0
+      test_decode "(-infinity,upper] - infinite lower bound", "'(,10]'::int4range", nil...11
+      test_decode "[lower,+infinity) - infinite upper bound", "'[1,)'::int4range", 1...nil
+      test_decode "(-infinity,+infinity) - both bounds infinite", "'(,)'::int4range", nil...nil
+    end
+
+    describe "int8range" do
+      test_decode "(lower,upper) - both exclusive", "'(3000000000,4000000000)'::int8range", 3_000_000_001...4_000_000_000
+      test_decode "(lower,upper] - exclusive lower, inclusive upper", "'(3000000000,4000000000]'::int8range", 3_000_000_001...4_000_000_001
+      test_decode "[lower,upper) - inclusive lower, exclusive upper", "'[3000000000,4000000000)'::int8range", 3_000_000_000...4_000_000_000
+      test_decode "[lower,upper] - both inclusive", "'[3000000000,4000000000]'::int8range", 3_000_000_000...4_000_000_001
+      test_decode "empty range", "'empty'::int8range", 0_i64...0_i64
+      test_decode "(-infinity,upper] - infinite lower bound", "'(,4000000000]'::int8range", nil...4_000_000_001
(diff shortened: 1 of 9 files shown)


spec/pg/decoders/range_decoder_spec.cr
```diff
@@ -0,0 +1,109 @@
...
+      test_decode "[lower,upper] - both inclusive", "'[2023-01-01,2023-12-31]'::daterange", Time.utc(2023, 1, 1)...Time.utc(2024, 1, 1)
+      test_decode "empty range", "'empty'::daterange", Time.unix(0)...Time.unix(0)
+      test_decode "(-infinity,upper] - infinite lower bound", "'(,2023-12-31]'::daterange", nil...Time.utc(2024, 1, 1)
+      test_decode "[lower,+infinity) - infinite upper bound", "'[2023-01-01,)'::daterange", Time.utc(2023, 1, 1)...nil
+      test_decode "(-infinity,+infinity) - both bounds infinite", "'(,)'::daterange", nil...nil
+    end
+
+    describe "tsrange" do
+      test_decode "(lower,upper) - both exclusive", "'(2023-01-01 10:30:00,2023-12-31 15:45:00)'::tsrange", Time.utc(2023, 1, 1, 10, 30, 0)...Time.utc(2023, 12, 31, 15, 45, 0)
+      test_decode "(lower,upper] - exclusive lower, inclusive upper", "'(2023-01-01 10:30:00,2023-12-31 15:45:00]'::tsrange", Time.utc(2023, 1, 1, 10, 30, 0)..Time.utc(2023, 12, 31, 15, 45, 0) # Crystal can't represent exclusive lower
```
straight-shoota: **issue:** Changing the exclusive lower to an inclusive one is a dangerous change.
Unfortunately, we cannot represent that with `Range`. But I'd consider silently converting the range to different semantics an error. 

For `Time` we can work around by treating it as a discrete type. Just add one nanosecond:
```suggestion
      test_decode "(lower,upper] - exclusive lower, inclusive upper", "'(2023-01-01 10:30:00,2023-12-31 15:45:00]'::tsrange", Time.utc(2023, 1, 1, 10, 30, 0, nanosecond: 1)..Time.utc(2023, 12, 31, 15, 45, 0)
```
This doesn't preserve the _exact_ semantics either, but it's much closer and preserves the _intent_.

And it probably won't work for `PG::Numeric`?

Alternatively, we need a different solution (e.g. a custom `Range` type).
Omitting this case (i.e. raise an error) would also be better than implicitly changing semantics.

spec/pg/decoders/range_decoder_spec.cr
```diff
@@ -0,0 +1,109 @@
+require "../../spec_helper"
+
+describe PG::Decoders do
+  describe "ranges" do
+    describe "int4range" do
+      test_decode "(lower,upper) - both exclusive", "'(1,10)'::int4range", 2...10 # PostgreSQL canonicalizes discrete ranges to [a,b) form
+      test_decode "(lower,upper] - exclusive lower, inclusive upper", "'(1,10]'::int4range", 2...11
+      test_decode "[lower,upper) - inclusive lower, exclusive upper", "'[1,10)'::int4range", 1...10
+      test_decode "[lower,upper] - both inclusive", "'[1,10]'::int4range", 1...11 # [a,b] becomes [a,b+1) for discrete types
```
straight-shoota, line 9: **question:** Why not use inclusive `Range` here?

spec/pg/decoders/range_decoder_spec.cr
mamantoha: Right. Silently converting exclusive lower bounds to inclusive ones changes the semantics and could lead to bugs.

https://www.postgresql.org/docs/current/rangetypes.html#RANGETYPES-DISCRETE

- Discrete types (`int4`, `int8`, `date`): have clear "next" values (`1 -> 2`,  `2023-01-01` -> `2023-01-02`). This is working.
- Continuous types (`timestamp`, `numeric`): don't have clear "next" values.

From the documentation:

> Even though timestamp has limited precision, and so could theoretically be treated as discrete, it's better to consider it continuous since the step size is normally not of interest.

Do you think I should implement these adjustments?

spec/pg/decoders/range_decoder_spec.cr
mamantoha: https://www.postgresql.org/docs/current/rangetypes.html#RANGETYPES-DISCRETE

> The built-in range types int4range, int8range, and daterange all use a canonical form that includes the lower bound and excludes the upper bound; that is, [). User-defined range types can use other conventions, however.

So when you write `'[1,10]'::int4range` in PostgreSQL:

1. You specify `[1,10]` (both inclusive)
2. PostgreSQL internally converts it to` [1,11)` (inclusive lower, exclusive upper)
3. The Crystal decoder receives this canonical `[1,11)` form
4. Crystal represents it as `1...11` (which is exactly `[1,11)`)

straight-shoota: Understood, this normalization comes from the server 👍

spec/pg/decoders/range_decoder_spec.cr
mamantoha: Instead of approximating timestamp ranges by adding a small increment, I changed the implementation to decode ranges into a custom `PG::Range(T)` type. This preserves the PostgreSQL lower/upper inclusivity flags exactly, including cases like `(a,b]`.

Native Crystal `Range` is still supported for parameter encoding as a convenience, but decoded values now use `PG::Range(T)` so continuous types like `tsrange`, `tstzrange`, and `numrange` do not lose boundary information.

I also updated the specs to assert exact exclusive-lower behavior instead of relying on native `Range`.

src/pg/decoders/range_decoder.cr
```diff
@@ -0,0 +1,234 @@
+module PG
+  module Decoders
+    # Range flags from PostgreSQL range format
+    RANGE_EMPTY  = 0x01
+    RANGE_LB_INC = 0x02 # Lower bound inclusive
+    RANGE_UB_INC = 0x04 # Upper bound inclusive
+    RANGE_LB_INF = 0x08 # Lower bound infinite
+    RANGE_UB_INF = 0x10 # Upper bound infinite
```
jgaskins: Would it make sense to put these into a mixin that only the `RangeDecoder` and `MultiRangeDecoder` types include?

src/pg/decoders/range_decoder.cr
```diff
@@ -0,0 +1,234 @@
...
+                  decode_element(io)
+                end
+
+        PG::Range.new(
+          lower,
+          upper,
+          lower_inclusive: lower_bound_inclusive,
+          upper_inclusive: upper_bound_inclusive
+        )
+      end
```
jgaskins: Same as above. Does this need to be in the mixin included by all decoders?

src/pg/decoders/range_decoder.cr
```diff
@@ -0,0 +1,234 @@
...
+        count.times do
+          # Each range element has a 4-byte length followed by the range data
+          read_i32(io)
+
+          range = decode_range(io, bytesize, oid)
+
+          ranges << range
+        end
+
+        ranges
```
jgaskins: The `Array` constructor takes a block, which simplifies this a bit.

```suggestion
        Array(PG::Range(T)).new(count) do
          # Each range element has a 4-byte length followed by the range data
          read_i32(io)

          decode_range(io, bytesize, oid)
        end
```

We probably don't need to pass the `bytesize` and `oid` to `decode_range`. The method doesn't use those values. If it did, they'd be incorrect at that point. For example, we'd be passing in the `oid` for an `int4multirange` but we need it to decode an `int4range`. We've also consumed at least 4 bytes from the `io` at this point, but `bytesize` would still represent the size of the entire `int4multirange` value rather than the `int4range` for it to read on every iteration of the block.

src/pg/decoder.cr
```diff
@@ -2,7 +2,7 @@ require "json"
 require "uuid"
 
 module PG
-  alias PGValue = String | Nil | Bool | Int32 | Float32 | Float64 | Time | JSON::Any | PG::Numeric | UUID
+  alias PGValue = String | Nil | Bool | Int32 | Float32 | Float64 | Time | JSON::Any | PG::Numeric | UUID | PG::Range(Int32) | PG::Range(Int64) | PG::Range(Time) | PG::Range(PG::Numeric) | Array(PG::Range(Int32)) | Array(PG::Range(Int64)) | Array(PG::Range(Time)) | Array(PG::Range(PG::Numeric))
```
jgaskins: [nitpick] This line is almost 300 characters wide with this change, so now might be a good time to wrap it for readability.

```suggestion
  alias PGValue = Union(
    String,
    Nil,
    Bool,
    Int32,
    Float32,
    Float64,
    Time,
    JSON::Any,
    PG::Numeric,
    UUID,
    PG::Range(Int32),
    PG::Range(Int64),
    PG::Range(Time),
    PG::Range(PG::Numeric),
    Array(PG::Range(Int32)),
    Array(PG::Range(Int64)),
    Array(PG::Range(Time)),
    Array(PG::Range(PG::Numeric)),
  )
```

I'm recommending the `Union(...)` format over `A | B | C` for two reasons:
1. The Crystal formatter's indentation when using the `A | B | C` syntax over multiple lines is not how I would write it
2. Using the `Union(...)` syntax allows us to use the trailing comma, avoiding changing lines unnecessarily in the future. Appending a type with the `A | B | C` syntax would require us to append a `|` to the end of a line that otherwise didn't need to change.

src/pg.cr
```diff
@@ -1,6 +1,17 @@
...
+module DB::MetadataValueConverter
+  # Log Range as string representation rather than array
+  def self.arg_to_log(arg : Range) : ::Log::Metadata::Value
+    ::Log::Metadata::Value.new(arg.to_s)
+  end
+
+  def self.arg_to_log(arg : PG::Range) : ::Log::Metadata::Value
+    ::Log::Metadata::Value.new(arg.to_s)
+  end
+end
```
jgaskins, line 14: Is this used anywhere?

src/pq/param.cr
```diff
@@ -77,6 +78,127 @@ module PQ
       text "#{val.months} months #{val.days} days #{val.microseconds} microseconds"
     end
 
+    private def self.format_time(value : Time)
+      Time::Format::RFC_3339.format(value, fraction_digits: 9)
```
jgaskins: I recognize that you were just moving this definition down here, but we can also simplify this.

```suggestion
      value.to_rfc3339(fraction_digits: 9)
```

I kiiiiinda wanna also reduce this to 6 fraction digits because Postgres only stores timestamps with microsecond precision, so it would remove 3 bytes from every timestamp being sent as a query param. I'm not 100% confident that there won't be issues with rounding/truncating, though, so I'm not going to recommend it here.

jgaskins (changes requested):
This is great. I occasionally work on things where `tstzrange`s would be helpful and it'd be nice to be able to use them without having to add support for them in the app.

My notes here are mostly cosmetic. The only structural things are the placement of the constants and the `decode_range` method.

src/pg.cr
mamantoha: https://github.com/crystal-lang/crystal-db/pull/223

mamantoha:
@jgaskins thanks for the review. I addressed the remaining comments:

- Moved the range flags and `decode_range` into a range-specific mixin included only by `RangeDecoder` and `MultiRangeDecoder`.
- Removed the unused `bytesize` and `oid` parameters from `decode_range`.
- Kept the `Range` metadata logging overload for compatibility with `db 0.14.0`, and added a comment explaining that it mirrors the upstream crystal-db fix until this shard can rely on that version.

src/pg.cr
jgaskins: Ooh, TIL

src/pg/range.cr
```diff
@@ -0,0 +1,60 @@
...
+      !@lower_inclusive
+    end
+
+    def excludes_end?
+      !@upper_inclusive
+    end
+
+    def to_crystal_range
+      ::Range.new(@begin, @end, exclusive: excludes_end?)
+    end
```
jgaskins, line 30: Silently dropping the `excludes_begin?` here may lead to some hard-to-pin-down bugs. I want to say this should raise in this case so folks know exactly where the problem is. That's easy for me to say as someone who only ever uses ranges that are closed on the left, though. I don't know of any use cases where the range is open on the left (and the normalization mentioned in [this comment](https://github.com/will/crystal-pg/pull/304#discussion_r2527442994) doesn't apply to `tstzrange`), so I'm open to alternatives.

jgaskins (commented):
I do have one additional note below that I noticed during testing and we should probably discuss it before merging, but otherwise, I really like this.

Benchmarks look great:

```crystal
require "../src/pg"
require "benchmark"

pg = DB.open("postgres:///")
sql = <<-SQL
  SELECT
    '[2026-06-28T12:00:00-04, 2026-06-28T13:00:00-04)'::tstzrange AS time
  FROM generate_series(1, $1)
SQL

# Warm the prepared statement cache
pg.query_all sql, 1, as: Event

event = nil
allocations = {} of Int64 => Int64
Benchmark.bm do |x|
  {
    1,
    10,
    100,
    1_000,
    10_000,
    100_000,
    1_000_000,
    10_000_000,
  }.each do |count|
    x.report "#{count.humanize(precision: 0, significant: false)} events" do
      mem = Benchmark.memory do
        pg.query_each sql, count do |rs|
          event = Event.new(rs)
        end
      end
      allocations[count] = mem
    end
  end
end
puts "Heap allocations:"
pp allocations
  .transform_keys(&.humanize(precision: 0, significant: false))
  .transform_values(&.humanize_bytes)
pp event

struct Event
  include DB::Serializable

  @[DB::Field(converter: TSTZRangeConverter(Time, Time))]
  getter time : Range(Time, Time)
end

module TSTZRangeConverter(Begin, End)
  def self.from_rs(rs : DB::ResultSet) : Range(Begin, End)
    range = rs.read(PG::Range).to_crystal_range

    Range.new(
      range.begin.as(Begin),
      range.end.as(End),
      exclusive: range.excludes_end?,
    )
  end
end
```

</details>

```
                  user     system      total        real
1 events      0.000006   0.000011   0.000017 (  0.000054)
10 events     0.000005   0.000009   0.000014 (  0.000042)
100 events    0.000018   0.000010   0.000028 (  0.000067)
1k events     0.000122   0.000018   0.000140 (  0.000264)
10k events    0.001157   0.000117   0.001274 (  0.001776)
100k events   0.010441   0.000912   0.011353 (  0.015458)
1M events     0.078240   0.008208   0.086448 (  0.138686)
10M events    0.802205   0.085781   0.887986 (  1.319421)
Heap allocations:
{"1" => "368B",
 "10" => "384B",
 "100" => "368B",
 "1k" => "368B",
 "10k" => "368B",
 "100k" => "368B",
 "1M" => "368B",
 "10M" => "352B"}
Event(
 @time=
  2026-06-28 12:00:00-04:00[America/New_York]...2026-06-28 13:00:00-04:00[America/New_York])
```

The result is constant heap usage, which I'd hoped for but am still delighted to see, and sublinear CPU time, which was pleasantly surprising.

The ergonomics of going through `PG::Range` aren't ideal, but the correctness is more important. I think we could follow up with something like [this](https://github.com/will/crystal-pg/blob/9ce74e0c72b93a7c4ceca8b2f6fee00038108c74/src/pg_ext/big_decimal.cr#L18-L26) to allow folks to avoid not only writing a converter but also having to use it in all their range-based column-mapped properties:

```crystal
struct Event
  include DB::Serializable

  getter time : Range(Time, Time)
end
```

src/pg/range.cr
mamantoha: What would you recommend here?

1. Make `to_crystal_range` raise when `excludes_begin?` is true.
2. Keep the current behavior, but document clearly that `to_crystal_range` only preserves the upper-bound exclusivity and cannot preserve exclusive lower bounds.
3. Remove `to_crystal_range` entirely and require callers to inspect/use `PG::Range` when exact semantics matter.

jgaskins: My gut feeling is that it should raise because it can't be represented in the stdlib `Range` type. I do like your second option, though. It's not as disruptive, but that can also lead to it being easily overlooked.

The third option doesn't seem viable. All of the Postgres value types that don't map cleanly to stdlib types have some way of converting to a stdlib type. For example, `PG::Numeric` can be converted to a [`BigDecimal`](https://github.com/will/crystal-pg/blob/9ce74e0c72b93a7c4ceca8b2f6fee00038108c74/src/pg_ext/big_decimal.cr#L7-L15) or [`BigRational`](https://github.com/will/crystal-pg/blob/9ce74e0c72b93a7c4ceca8b2f6fee00038108c74/src/pg_ext/big_rational.cr#L7-L15), `PG::Interval` can be converted to [`Time::Span` or `Time::MonthSpan`](https://github.com/will/crystal-pg/blob/9ce74e0c72b93a7c4ceca8b2f6fee00038108c74/src/pg/interval.cr#L13-L33), etc. This is important because the data objects in this shard aren't full value types (there's no `Time#+(PG::Interval)`). They're just bags of metadata about the value we got back from Postgres.

`PG::Interval` is a good example of your second option. It doesn't map losslessly to either `Time::Span`, `Time::MonthSpan`, or even both combined since converting 1 calendar day to 24 monotonic hours is a lossy conversion[^1]. But it does it because, at least at the time it was written, there wasn't a better option for that. Maybe that's the path we take for `PG::Range`, too. Thoughts?

[^1]: Not all days are exactly 24 hours long ([example](https://github.com/jgaskins/duration#calendar-days)). Part of the reason I created [my `Duration` shard](https://github.com/jgaskins/duration) and [added Postgres support](https://github.com/jgaskins/duration#using-with-postgres) was to cover this gap. It maps better to the `PG::Interval` type than the stdlib types do. It wasn't _just_ for Postgres, but it exists for the same reason Postgres stores intervals the way it does.

jgaskins: @mamantoha Do you have any thoughts on what I wrote above?

Issue #307, opened 2026-06-26
Road to 1.0: comprehensive audit + offer to contribute fixes


n-rodriguez:
Hi 👋

While evaluating `crystal-pg` for a Crystal project, I ran a full multi-dimension audit toward a `1.0` tag (correctness, concurrency / fiber-safety, error model, perf/memory/CPU, PostgreSQL protocol completeness, public API / SemVer surface, tests / CI). I want to lead with the honest part: **the architecture is genuinely good.** A clean two-layer split — the wire protocol in `PQ::*` cleanly separated from the `crystal-db` driver in `PG::*` — correct binary decoders for the large majority of types, **SCRAM-SHA-256 + channel binding** (rare and well done), logical replication (`pgoutput`), and a real round-trip spec suite against Postgres 14–18. The bones are solid.

This is meant as a constructive map of what stands between today and a `1.0` that can be promised — and **I'm offering to contribute the fixes**, starting with the blockers and the most-requested security/recovery gaps.

📄 **Full audit (canonical, cited `file:line`):** https://gist.github.com/n-rodriguez/4cbd8a46b0892adc8305156c70273a50

## Summary by tier

| Tier | Theme | Headline |
|------|-------|----------|
| 0 | **Blockers** | `PG::Interval#to_spans` calls methods that don't exist (won't compile when called); `.size` instead of `.bytesize` corrupts the startup/password packet for non-ASCII; a stray `pp` debug print in replication; CI triggers on `main` while the default branch is `master` (push CI never fires); a decode error poisons a pooled connection silently |
| 1 | **Security** | `verify-ca`/`verify-full` are accepted but the TLS context is hard-coded `VerifyMode::NONE` — no cert/hostname verification (silent MITM exposure); an over-permissive `.pgpass` is warned about but read anyway; `PGSSLMODE`/`PGSSL*` env vars are ignored |
| 2 | **Correctness & concurrency** | RowDescription `table_oid`/column fields swapped + truncated; `infinity`/`-infinity` timestamps unhandled; `Numeric#to_s` fractional truncation is fragile; two array code paths disagree on non-1 lower bounds; the LISTEN/NOTIFY loop shares the socket read path with queries without a lock; `@@location_cache` is mutated cross-fiber without synchronization; two disjoint exception trees (`PQ::PQError` is not under `PG::Error`) |
[...]

Issue #196, opened 2019-11-23
Connection pool unreliable?


m-o-e:
Currently `DB::Database` does not recover from temporary
connection failures. It does not seem to evict broken connections
from the pool but continues to attempt to use them.

This means just a single brief outage will
leave the pool in a permanently broken state.

The only way to recover from connection failures appears to
be to discard the entire `DB::Database` instance and create a new pool.

Is this a bug or am I using it wrong? 🤔 

-----

### Test-case:

```
require "db"
require "pg"

initial_pool_size=5
max_pool_size=5
retry_attempts=99999

url = "postgres://localhost/test?initial_pool_size=#{initial_pool_size}&max_pool_size=#{max_pool_size}&max_idle_pool_size=#{max_pool_size}&checkout_timeout=5.0&retry_attempts=#{retry_attempts}&retry_delay=1"

puts "Trying: #{url}"
db = DB.open(url)
loop do
  begin
    db.query("select 42").close
    puts "."
  rescue e
    puts e
  end
  sleep 1
end
```

Just leave the above running and while it runs
restart your local postgres server.

### Regardless of what settings you choose, you will see the following output:

```
.
.
.
Error reading socket: Connection reset by peer
Error writing to socket: Broken pipe
Error writing to socket: Broken pipe
Error writing to socket: Broken pipe
[...]
```

bcardiff:
As long as you are not checking out connections explicitly the pool should be able to recover from temporary outages. Sample in mysql https://crystal-lang.org/reference/database/connection_pool.html

Maybe an exception with the wrong type is raised and the connection loss is not detected.

But you snippets looks good and i would expect to be resilient.

m-o-e:
Possibly it's postgres specific?

I couldn't find a way to make my above snippet resilient - it always breaks
when the server is restarted.

straight-shoota:
Yeah, I've encountered similar errors with postgres as well, but never did any deeper investigation. So maybe it's an issue in `crystal-pg`.

stakach:
also finding this an issue working with GCP's hosted postgres compatible database and new connections take upwards of 2 seconds to connect meaning we really need to rely on the pool.
So we have the pool configured with `max_idle_pool_size` == `max_pool_size` == `50` but if nothing is going on for awhile the connections eventually time out and broken sockets are returned causing 500 errors in our web app `DB::ConnectionLost` errors

Also seeing the pool not recovering from DB outages - we work around that using health checks, but it's not optimal

[...]

PR #308, opened 2026-06-26
Tier 0: fix mechanical blockers — Interval#to_spans, startup/password bytesize, replication debug noise, CI branch

n-rodriguez:
Part of the road-to-1.0 audit in #307. This is **PR 1 of the Tier 0 work** — the mechanical blockers: things that pass `crystal build` but break when the path is exercised, or silently never run. Each fix lands with a failing-first spec.

## Fixes

- **`PG::Interval#to_spans` referenced undefined methods** (`to_time_span` / `to_time_month_span`) and could not compile the moment it was called. It now returns `{to_span(approx_months: 0), to_month_span}` — a lossless decomposition (time part → `Time::Span`, months → `Time::MonthSpan`, no month→day approximation), pinned by a round-trip spec (`2024-01-31 + 2mo + 3d + 4h == 2024-04-03 04:00:00`).
- **Startup & password length prefixes used `String#size` instead of `#bytesize`** (`src/pq/connection.cr`), under-counting the length for any non-ASCII `user` / `database` / `application_name` / `password` and corrupting the packet. Two one-line fixes + a multi-byte `application_name` spec.
- **Replication debug noise**: removed a stray `pp size: size` in `error_frame.cr` (it printed on every error frame) and a dead `IO::Memory` allocation in `copy_data.cr`. Added the first unit spec for `ErrorFrame` decoding.
- **CI never ran on pushes**: the workflow triggered on `main`, but the default branch is `master`, so push builds never fired. Changed to `master`.

## Notes

- **Scope**: strictly these four blockers. Blocker 0.5 from the audit (a decode error can leave a pooled connection in a desynced wire state — related to #196) is intentionally deferred to a follow-up PR, since it changes connection-state handling and deserves its own discussion.
- **Test assertion detail**: the multi-byte `application_name` spec asserts the connection is *accepted* rather than reading the value back. Under a C-locale UTF8 server — which is what the test setup (`flake.nix` → `initdb --no-locale --encoding=UTF8`) produces — PostgreSQL byte-escapes non-ASCII GUC output on every read path (`SHOW` / `current_setting` / `pg_stat_activity`), so a value round-trip can't be asserted portably. Without the fix the malformed startup packet is rejected (connection refused), so connection-acceptance is the real discriminating signal.
- Full spec suite green locally, including the auth and replication specs.

> ℹ️ Transparency: these fixes came out of an audit produced with Claude Opus 4.8 [1M]; the compile/wire-format findings and each fix were verified by hand against the source and behind specs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,7 +1,7 @@
 on:
   push:
     branches:
-      - main
+      - master
   pull_request: {}
   workflow_dispatch: {}
 jobs:
--- a/spec/pg/interval_spec.cr
+++ b/spec/pg/interval_spec.cr
@@ -52,4 +52,24 @@ describe PG::Interval do
       interval.to_month_span.should eq(Time::MonthSpan.new(123))
     end
   end
+
+  describe "to_spans" do
+    it "splits an interval with months into a Time::Span and a Time::MonthSpan" do
+      # 2 months, 3 days, 4 hours
+      interval = PG::Interval.new(microseconds: 4_i64 * 3600 * 1_000_000, days: 3, months: 2)
+      interval.to_spans.should eq({Time::Span.new(days: 3, hours: 4), Time::MonthSpan.new(2)})
+    end
+
+    it "round-trips through a Time" do
+      interval = PG::Interval.new(microseconds: 4_i64 * 3600 * 1_000_000, days: 3, months: 2)
+      span, month_span = interval.to_spans
+      base = Time.utc(2024, 1, 31, 0, 0, 0)
+      (base + span + month_span).should eq(Time.utc(2024, 4, 3, 4, 0, 0))
+    end
+
+    it "works for an interval without months" do
+      interval = PG::Interval.new(microseconds: 0, days: 5, months: 0)
+      interval.to_spans.should eq({Time::Span.new(days: 5), Time::MonthSpan.new(0)})
+    end
+  end
 end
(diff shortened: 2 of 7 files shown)


n-rodriguez:
ping @will @jgaskins

n-rodriguez:
Rebased onto the latest `master`.

Two of the changes this PR originally carried — removing the `pp` debug output in `error_frame.cr` and the unused `IO::Memory` in `copy_data.cr` — landed directly on `master` in the meantime (e28494e, 9ce74e0), so they've dropped out of the diff here. 👍

What remains in this PR:

- **`PG::Interval#to_spans`** — it referenced undefined methods (`to_time_span`/`to_time_month_span`) and couldn't compile when called; now returns `{to_span(approx_months: 0), to_month_span}` with a round-trip spec.
- **`.size` → `.bytesize`** for the startup and password length prefixes (`pq/connection.cr`) — the character count under-counts the wire length for any non-ASCII user/database/application_name/password; covered by a multi-byte `application_name` spec.
- **First unit spec for `ErrorFrame` decoding** — the `error_frame.cr` change here is now just tightening the (now-unused) `size = read(io, Int32)` binding to a bare `read(io, Int32)`.
- **CI** — run push builds on `master` (the actual default branch) instead of `main`, which never fired.

Rebased cleanly; `crystal build --error-on-warnings` and `crystal tool format --check` pass. Happy to split any of these into separate PRs if that's easier to review.

Issue #237, opened 2021-11-10
cockroach db


wontruefree:
I am having trouble connecting to the free instance of coackroachdb.  I have read through the issues, and it seems to be connected to certs.  I have tried disabling SSL and not including any SSL options.

It seems like there are two problems.  SSL certs and options params.  I am not sure I can get the client to get the correct SSL cert to use to connect to cockroachdb.  Also, options have to be passed to cockroachdb to enable a lot of the features, including clusters.  from the return errors. I do not think the options are being passed correctly.

here is a connection url you can get from the cockroachdb website
postgresql://user:pass@domain.cockroachlabs.cloud:26257/defaultdb?sslmode=verify-full&sslrootcert=$HOME/.postgresql/root.crt&options=--cluster%3Dkcrystal-1234

will:
The options params at least I don’t think is a general postgres protocol thing. libpq does a few to manage how it operates but it's all over the place what libpq does with them. Some change what types of SCRAM is accepted, some say where to look for certs. So I'm not sure what a `options=--cluster%3Dkcrystal-1234` would actually do?

cyangle:
@wontruefree You should be able to connect with below connection string, you need to add the cluster name as db prefix and accept cleartext auth method.

But the problem is that it seems like the option `sslmode=verify-full` has no effect, it still connects without providing the `sslrootcert`.

`postgresql://user:pass@domain.cockroachlabs.cloud:26257/crystal-1234.defaultdb?auth_methods=cleartext,md5,scram-sha-256&sslmode=verify-full&sslrootcert=$HOME/.postgresql/root.crt`

@will Does `crystal-pg` support `sslmode=verify-full`?

Issue #231, opened 2021-07-01
Unable to connect to Cockroachdb


rishavs:
I am having issues connecting to cockroachdb and I wonder if this is the same issue.

my connection string is something like;
`postgres://uname@somehosturl.cockroachlabs.cloud:26257/defaultdb?sslmode=verify-full&sslrootcert=cc-ca.crt&options=--cluster=somedbname`

```
Unhandled exception:  (DB::ConnectionRefused)
  from lib/db/src/db/connection.cr:34:0 in 'initialize'
  from lib/pg/src/pg/connection.cr:7:5 in 'new'
  from lib/pg/src/pg/driver.cr:3:5 in 'build_connection'
  from lib/db/src/db/database.cr:57:9 in '->'
  from /usr/share/crystal/src/primitives.cr:255:3 in 'build_resource'
  from lib/db/src/db/pool.cr:47:34 in 'initialize'
  from lib/db/src/db/pool.cr:40:5 in 'new:initial_pool_size:max_pool_size:max_idle_pool_size:checkout_timeout:retry_attempts:retry_delay'
  from lib/db/src/db/database.cr:56:15 in 'initialize'
  from lib/db/src/db/database.cr:49:5 in 'new'
  from lib/db/src/db.cr:155:5 in 'build_database'
  from lib/db/src/db.cr:151:5 in 'build_database'
  from lib/db/src/db.cr:119:5 in 'open'
  from src/digglu.cr:22:5 in '__crystal_main'
  from /usr/share/crystal/src/crystal/main.cr:110:5 in 'main_user_code'
  from /usr/share/crystal/src/crystal/main.cr:96:7 in 'main'
  from /usr/share/crystal/src/crystal/main.cr:119:3 in 'main'
  from __libc_start_main
  from _start
  from ???
Caused by: CodeParamsRoutingFailed: rejected by BackendConfigFromParams: Invalid cluster name (PQ::PQError)
  from lib/pg/src/pq/connection.cr:203:7 in 'handle_error'
  from lib/pg/src/pq/connection.cr:186:7 in 'handle_async_frames'
  from lib/pg/src/pq/connection.cr:162:7 in 'read'
  from lib/pg/src/pq/connection.cr:157:7 in 'read'
  from lib/pg/src/pq/connection.cr:414:31 in 'expect_frame'
  from lib/pg/src/pq/connection.cr:413:5 in 'expect_frame'
  from lib/pg/src/pq/connection.cr:242:20 in 'connect'
[...]

PR #309, opened 2026-06-26
Implement sslmode=verify-ca and verify-full (real TLS certificate verification)

### LINKED ISSUE #240: Proper implementation of `sslmode=verify-full`
Given a lot of people use `cleartext` auth method to connect to DBs hosted by Azure and CockroachLabs.

I think it's time to implement `sslmode=verify-full` to protect these use cases.

n-rodriguez:
Part of the road-to-1.0 audit in #307 (Tier 1, security). Closes #240; helps #237 / #231.

Previously `sslmode=verify-ca` and `sslmode=verify-full` were *accepted* but the TLS context was hard-coded to `OpenSSL::SSL::VerifyMode::NONE` — so a user asking for `verify-full` got an **unauthenticated** TLS channel with no certificate or hostname verification (a silent MITM exposure). This implements real verification for both modes.

## What changed

- **`verify-full`** → `VerifyMode::PEER` + CA, and the socket is created with `hostname:`, so OpenSSL enforces both the certificate chain and the hostname/IP.
- **`verify-ca`** → verifies the certificate chain only (no hostname), via a custom cert-verify callback that clears the host check, **while still sending SNI** (the socket still gets `hostname:`).
- **`disable` / `allow` / `prefer` / `require`** → unchanged (`NONE`); the existing `require` "server must offer SSL" backstop is kept.
- **Root certificate resolution** (`PQ::ConnInfo#ssl_root_cert`): explicit `sslrootcert` → `PGSSLROOTCERT` → `~/.postgresql/root.crt` → the system CA store (already loaded by `OpenSSL::SSL::Context::Client`). So `verify-full` works out of the box against servers with a publicly-trusted cert, and with a private CA when one is provided.

## Tests

New `spec/pq/ssl_verify_spec.cr` (gated on `CRYSTAL_PG_CERT_DIR`, like the existing auth specs): for both `verify-ca` and `verify-full`, a trusted CA connects and an **untrusted CA is rejected** (`DB::ConnectionRefused`) — i.e. verification genuinely happens. `flake.nix`'s test server cert gains `subjectAltName=IP:127.0.0.1` so `verify-full` can be exercised against `127.0.0.1`. Unit tests cover the root-cert resolution order. Full suite green.

## Known limitations (documented in-code)

- **`verify-ca` to an IP literal** that is *not* in the server cert's SAN is still rejected (the callback clears the hostname check but not the IP check — clearing the IP param via a NULL argument hangs the handshake). This is **fail-closed** (stricter than libpq, no security weakening) and a rare case; `verify-ca` to a hostname, or to a matching IP, works.
- A couple of negative cases (hostname-mismatch rejection; the `verify-ca` IP-skip) are not asserted in the suite because they require a loopback name/address that the driver currently can't connect to without hanging (a separate connect-timeout / IPv6-host issue, out of scope here).

Scope is strictly server-certificate verification. Client-certificate auth (the pending `ssl clientcert` spec) is intentionally left to a follow-up, though the SAN cert added here is a step toward enabling it.

> ℹ️ Transparency: produced with Claude Opus 4.8 [1M]; the OpenSSL/TLS behavior was verified by hand against the Crystal stdlib and exercised behind specs against a CI-faithful local server.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -23,7 +23,8 @@
             openssl req -new -nodes -text -out ca.csr -keyout ca-key.pem -subj "/CN=certificate-authority"
             openssl x509 -req -in ca.csr -text -signkey ca-key.pem -out ca-cert.pem
             openssl req -new -nodes -text -out server.csr -keyout server-key.pem -subj "/CN=pg-server"
-            openssl x509 -req -in server.csr -text -CA ca-cert.pem -CAkey ca-key.pem -CAcreateserial -out server-cert.pem
+            echo "subjectAltName=IP:127.0.0.1" > san.ext
+            openssl x509 -req -in server.csr -text -CA ca-cert.pem -CAkey ca-key.pem -CAcreateserial -extfile san.ext -out server-cert.pem
             openssl req -new -nodes -text -out client.csr -keyout client-key.pem -subj "/CN=crystal_ssl"
             openssl x509 -req -in client.csr -text -CA ca-cert.pem -CAkey ca-key.pem -CAcreateserial -out client-cert.pem
 
--- a/spec/pq/conninfo_spec.cr
+++ b/spec/pq/conninfo_spec.cr
@@ -144,3 +144,25 @@ describe PQ::ConnInfo, ".from_conninfo_string" do
     end
   end
 end
+
+describe "#ssl_root_cert" do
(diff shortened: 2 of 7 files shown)


n-rodriguez:
ping @will @jgaskins

PR #311, opened 2026-09-08
Implement SSL verify-full and verify-ca

plambert:
Changes `sslmode=verify-full` and `sslmode=verify-ca` to no longer accept plaintext connections--only `sslmode-require` did so before.

Adds a method `#ssl_required?` to answer whether SSL must be negotiated.

Adds a method `#resolved_sslrootcert` to get the path to the SSL root certificate; this is determined from `$PGSSLROOTCERT` and/or the `sslrootcert` parameter, and defaults to `~/.postgres/root.crt` on Linux/Darwin, or `%APPDATA%/postgresql/root.crt` on Windows.

### DIFF
--- a/src/ext/openssl.cr
+++ b/src/ext/openssl.cr
@@ -82,3 +82,10 @@ end
     fun x509_digest = X509_digest(x509 : X509, evp_md : EVP_MD, hash : UInt8*, len : Int32*) : Int32
   end
 {% end %}
+
+# Needed to be able to support the ssl modes (verify-full, verify-ca, etc)
+{% unless LibCrypto.has_method?(:x509_store_ctx_get_error) %}
+  lib LibCrypto
+    fun x509_store_ctx_get_error = X509_STORE_CTX_get_error(ctx : X509_STORE_CTX) : Int32
+  end
+{% end %}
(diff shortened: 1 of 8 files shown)


src/ext/openssl.cr
```diff
@@ -82,3 +82,10 @@ end
     fun x509_digest = X509_digest(x509 : X509, evp_md : EVP_MD, hash : UInt8*, len : Int32*) : Int32
   end
 {% end %}
+
+# Needed to be able to support the ssl modes (verify-full, verify-ca, etc)
+{% unless LibCrypto.has_method?(:x509_store_ctx_get_error) %}
+  lib LibCrypto
+    fun x509_store_ctx_get_error = X509_STORE_CTX_get_error(ctx : X509_STORE_CTX) : Int32
+  end
+{% end %}
```
jgaskins: The conditional seems like it might avoid collision with [the same function declaration in the `spider-gazelle/openssl_ext` shard](https://github.com/spider-gazelle/openssl_ext/blob/b7c9ea5b1b79aa81fa014c05d0a21054ca2437fc/src/openssl_ext/lib_crypto.cr#L400), but from what I can tell, it's fine to redefine the same C function the exact same way. For example, this compiles and runs just fine:

```crystal
require "openssl"

lib LibCrypto
  fun err_get_error = ERR_get_error : ULong
end

lib LibCrypto
  fun err_get_error = ERR_get_error : ULong
end

pp LibCrypto.err_get_error
```

